PR #2, opened 2019-01-08
Support for Bitbucket Server downloading

ametzger:
Adds support for downloading users, projects, repositories, branches, commits, and pull requests for on-premise Bitbucket servers.

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -2,62 +2,132 @@ import argparse
 import os
 import sys
 import json
+import urllib3
 
 from jira import JIRA
 from jira.resources import GreenHopperResource
-
 from jf_agent.jira_download import download_users, download_fields, download_resolutions, download_issuetypes, \
     download_issuelinktypes, download_priorities, download_projects_and_versions, download_boards_and_sprints, \
     download_issues, download_worklogs
 
+import stashy
+from jf_agent.bb_download import get_all_users, get_all_projects, get_all_repos, get_default_branch_commits, get_pull_requests
+
+
 def main():
     parser = argparse.ArgumentParser()
-    parser.add_argument('jira_url', help='the fully-qualified base URL for your Jira server, e.g. https://jira.yourcompany.com')
-    parser.add_argument('-o', '--out_dir', nargs='?', default='.', help='output directory')
+    parser.add_argument(
+        '-o',
+        '--out_dir',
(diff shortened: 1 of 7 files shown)


mattklein:
Meta comment, applies to both this and the Jira agent code (@eliwind) -- it's a bit concerning that the jf_agent repo is introducing code that's essentially duplicated in the jellyfish repo. At some point it's going to be painful and error-prone for us to be maintaining both copies of code, there might be subtle differences, we'll forget to fix stuff in both places, etc. What'd be the path to unify them? Make jf_agent a dependency in jellyfish, and refactor the adapter "pull" code to use common code from jf_agent?

jf_agent/main.py
```diff
@@ -2,49 +2,58 @@
...
+    parser.add_argument(
+        '--jira-url',
+        type=str,
+        help='the fully-qualified base URL for your Jira server, e.g. https://jira.yourcompany.com',
+        required=False)
+    parser.add_argument(
+        '--bitbucket-url',
+        type=str,
+        help='the fully-qualified base URL for your Bitbucket server, e.g. https://bitbucket.yourcompany.com',
+        required=False)
```
eliwind: I feel like we should support BB cloud, GH cloud, and GHE too.  Fine to start with this one, but might want to think how the configuration works in general.

eliwind: the rationale for cloud providers being, they may still want to audit what info we're getting, and especially if we have a model where we ever read a local repo clone, it'll have to be agent-based.

jf_agent/main.py
```diff
@@ -2,49 +2,58 @@
...
-    jira_url = args.jira_url
     outdir = args.out_dir
-    
-    if jira_username and jira_password:
-        jira_connection = get_basic_jira_connection(jira_url, jira_username, jira_password)
-    else:
-        print('Jira credentials not found. Set environment variables, either JIRA_SHARED_SECRET, JIRA_CLIENT_KEY, JIRA_KEY (for OAuth) or JIRA_USERNAME and JIRA_PASSWORD (for basic auth)')
+    jira_url = args.jira_url
+    bb_url = args.bitbucket_url
+    skip_ssl_verification = args.no_verify_ssl
```
eliwind: Seems weird to expose this knob for bb but not jira

PR #6, opened 2019-03-08
Filter projects and fields

mattklein:


jf_agent/jira_download.py
```diff
@@ -40,11 +40,28 @@ def download_users(jira_connection):
...
+    fields = []
+    for field in jira_connection.fields():
+        keep = True
+        for filt in filters:
+            if not filt(field):
+                keep = False
+                break
+        if keep:
+            fields.append(field)
+
```
eliwind: I would have done that with a list comprehension:

`fields = [field for field in jira_connection.fields() if all(filt(field) for filt in filters)]`

PR #8, opened 2019-03-18
OJ-3334: Support for GDPR API changes

eliwind:


jf_agent/jira_download.py
```diff
@@ -271,13 +261,63 @@ def download_worklogs(jira_connection, issue_ids):
...
+        # jira_connection.search_users creates a query that is no longer accepted on
+        # GDPR-compliant Jira instances.  Construct the right query by hand
+        params = {
+            'startAt': start_at,
+            'maxResults': max_results,
+            'query': query,
+            'includeActive': include_active,
+            'includeInactive': include_inactive
+        }
+        return jira_connection._get_json('user/search', params)
```
ametzger: This seems like a nice thing we could push upstream, see [`search_users` upstream](https://github.com/pycontribs/jira/blob/f49b92e8f0482cf987ff6ed02834632d28fed4c8/jira/client.py#L2726-L2748) which looks quite similar. It looks like there is [an upstream bug](https://github.com/pycontribs/jira/issues/697) for dealing with this.

PR #9, opened 2019-05-07
handle empty repo

eliwind:
... and also blacken the code

### DIFF
--- a/jf_agent/bb_download.py
+++ b/jf_agent/bb_download.py
@@ -9,132 +9,121 @@ def datetime_from_bitbucket_server_timestamp(bb_server_timestamp_str):
 
 def _normalize_user(user):
     return {
-        "id": user["id"],
-        "login": user["name"],
-        "name": user["displayName"],
-        "email": user["emailAddress"],
+        'id': user['id'],
+        'login': user['name'],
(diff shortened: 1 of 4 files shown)


eliwind:
CircleCI is not properly configured for this repo -- disregard the failing check

jf_agent/bb_download.py
```diff
@@ -88,23 +90,22 @@ def get_default_branch_commits(client, repos):
...
-            commits = api_project.repos[repo['name']].commits(
-                until=repo['default_branch_name'])
+            commits = api_project.repos[repo['name']].commits(until=repo['default_branch_name'])
+
+            for commit in tqdm(
+                commits,
+                desc=f'downloading {repo["project"]["login"]}/{repo["name"]} commits',
+                unit='commit',
+            ):
+                yield _normalize_commit(commit, repo)
```
ametzger: @eliwind to confirm, this is the only non-formatting change?

eliwind: That's right

eliwind: the problem was that we had a try/catch around the thing that returns a generator of commits, but the NotFound wasn't actually thrown until you iterator over them

PR #12, opened 2019-05-23
Allow redacting names and URLs

eliwind:
Also fixes a bug with the --since handling; we were stopping pulling entirely as soon as we hit a single repo with a commit or PR before the since date instead of going on the next repo.  That is now fixed.

### DIFF
--- a/jf_agent/bb_download.py
+++ b/jf_agent/bb_download.py
@@ -1,8 +1,26 @@
 from datetime import datetime
-from tqdm import tqdm
 import stashy
 import re
 import pytz
+
+
+class NameRedactor:
+    def __init__(self):
+        self.redacted_names = {}
+        self.seq = 0
+
+    def redact_name(self, name):
+        redacted_name = self.redacted_names.get(name)
+        if not redacted_name:
+            redacted_name = f'redacted-{self.seq}'
+            self.seq += 1
+            self.redacted_names[name] = redacted_name
+        return redacted_name
+
+
+_branch_redactor = NameRedactor()
+_project_redactor = NameRedactor()
+_repo_redactor = NameRedactor()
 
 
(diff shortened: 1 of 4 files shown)


jf_agent/bb_download.py
```diff
@@ -19,34 +37,82 @@ def _normalize_user(user):
...
+def _normalize_project(api_project, redact_names_and_urls):
+    return {
+        'id': api_project['id'],
+        'login': api_project['key'],
+        'name': (
+            api_project.get('name')
+            if not redact_names_and_urls
+            else _project_redactor.redact_name(api_project.get('name'))
+        ),
+        'url': api_project['links']['self'][0]['href'] if not redact_names_and_urls else None,
```
jnicoleoliveira: Why does the absence of `readact_name_and_urls` return None in the url case, and the un-redacted(?) string in the name case?

jf_agent/bb_download.py
```diff
@@ -19,34 +37,82 @@ def _normalize_user(user):
...
+        (p, _normalize_project(p, redact_names_and_urls))
         for p in client.projects.list()
         if all(filt(p) for filt in filters)
     ]
+    print('✓')
+    return projects
+
+
+def _normalize_repo(api_project, api_repo, redact_names_and_urls):
+    repo = api_repo.get()
```
jnicoleoliveira: Some of the getter logic matches what we do in `adapters.bitbucket`. I wonder if it's worth factoring it out to avoid needing to update two places in the future.

jf_agent/bb_download.py
eliwind: Because names are just labels, and we can show "Redacted-0004" in the app, where the URL is used for generating a link, and so the app needs to know that there is no link (vs generating a link to a URL that doesn't actually exist).

PR #13, opened 2019-06-04
Retry on Connection and Authentication Failures

jnicoleoliveira:
Retry on authentication and connection failures by intercepting requests made from stashy module.

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -5,12 +5,12 @@ import json
 import urllib3
 import yaml
 import dateparser
-from datetime import datetime
 import pytz
+from datetime import datetime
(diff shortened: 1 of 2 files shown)


jf_agent/main.py
```diff
@@ -5,19 +5,20 @@
 import urllib3
 import yaml
 import dateparser
-from datetime import datetime
 import pytz
+from datetime import datetime
```
eliwind: unused?

jnicoleoliveira: Ah, the diff is misleading here. I just moved the import line, this is referenced in `line 76`

PR #16, opened 2019-06-07
Reset client on 401

eliwind:


jf_agent/bb_download.py
```diff
@@ -8,13 +8,51 @@
...
+        response = super().request(method, url, **kwargs)
+        if response.status_code == 401:
+            print(f'WARN: received 401 for the request [{method}] {url} - resetting client session')
+            self._client._client = stashy.client.StashClient(
+                self.base_url,
+                username=self.username,
+                password=self.password,
+                verify=self.verify,
+                session=self,
+            )
```
ametzger: Should this use `self.reset_client_session()`?

eliwind: whoops

jf_agent/bb_download.py
```diff
@@ -8,13 +8,30 @@
...
-def get_session():
+class StashySession(requests.Session):
+    def __init__(self, base_url):
+        self.base_url = base_url
+        super().__init__()
+
+    def request(self, method, url, **kwargs):
+        # If we get HTTP 401, re-authenticate and try again
+        response = super().request(method, url, **kwargs)
+        if response.status_code == 401 and method != 'HEAD':
```
ametzger: turns out [`requests.head` uses lowercase for this](https://github.com/kennethreitz/requests/blob/4983a9bde39c6320aa4f3e34e50dac6e263dab6f/requests/api.py#L101), I'd ignore case here

PR #18, opened 2019-06-25
OJ-4141: Add github support

eliwind:


jf_agent/main.py
```diff
@@ -68,9 +76,11 @@ def main():
 
     jira_config = config.get('jira', {})
     bb_config = config.get('bitbucket', {})
+    git_config = config.get('git', bb_config)
```
mattklein: Looks a little weird. Is it your thinking that we still want to support the `bitbucket` config for backwards compatibility? If so then maybe it'd be a bit clearer to single-line it as:
```
git_config = config.get('git', config.get('bitbucket', {}))
```

setup.py
```diff
@@ -5,13 +5,21 @@
...
     packages=setuptools.find_packages(),
-    install_requires=["jira", "tqdm", "stashy", "pyyaml>=4.2b1", "urllib3==1.24.2", "dateparser"],
+    install_requires=[
+        "jira",
+        "tqdm",
+        "stashy",
+        "pyyaml>=4.2b1",
+        "urllib3==1.24.2",
+        "dateparser",
+        "pygithub @ git+ssh://git@github.com/Jellyfish-AI/PyGithub@ed5ec0ab#egg=pygithub-1.43.2",
```
mattklein: I don't follow the syntax here -- what's ed5ec0ab? I don't see that hash in our fork of the repo. And what does it mean to have the #egg tag?

jf_agent/main.py
eliwind: Yeah, I was trying to be backwards-compatible.  I can one-line it if you think that's clearer.

setup.py
eliwind: https://python-packaging.readthedocs.io/en/latest/dependencies.html#packages-not-on-pypi

eliwind: That hash seems to be the commit on my branch, not the commit as merged into master.  I'll change this to the master one for clarity.

PR #19, opened 2019-06-27
GitHub support: remove dependency on pygithub

eliwind:


jf_agent/github_client.py
```diff
@@ -0,0 +1,115 @@
...
+        url = f'{self.base_url}/repos/{full_repo}/pulls?state=all&sort=updated&direction=desc'
+        return (self.get_json(m['url']) for m in self.get_all_pages(url))
+
+    def get_pr_comments(self, full_repo, pr_id):
+        url = f'{self.base_url}/repos/{full_repo}/pulls/{pr_id}/comments'
+        return self.get_all_pages(url)
+
+    def get_pr_reviews(self, full_repo, pr_id):
+        url = f'{self.base_url}/repos/{full_repo}/pulls/{pr_id}/reviews'
+        return self.get_all_pages(url)  # no rate limiting
```
mattklein: Misplaced/irrelevant comment?

jf_agent/github_client.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def get_pr_commits(self, full_repo, pr_id):
+        url = f'{self.base_url}/repos/{full_repo}/pulls/{pr_id}/commits'
+        return self.get_all_pages(url)  # no rate limiting
```
mattklein: Misplaced/irrelevant comment?

jf_agent/github_client.py
```diff
@@ -0,0 +1,115 @@
...
+                    logger.warn(
+                        f'Github rate limit exceeded.  Trying again in {str(reset_wait)}...'
+                    )
+                    time.sleep(reset_wait.seconds)
+                    continue
+
+                raise
+
+    # Handle pagination
+    def get_all_pages(self, url):
```
mattklein: Nice

setup.py
```diff
@@ -5,21 +5,13 @@
 
 setuptools.setup(
     name="jf_agent",
-    version="0.0.21",
+    version="0.0.20",
```
mattklein: You meaning to decrement the version?

eliwind: weirdly, yes: we apparently never actually released a .20, so we bumped it to .21 before unnecessarily.

eliwind:
@jnicoleoliveira Yeah, fair point; as far as I can tell, the API really does return all the info all the time, but I guess we'll see.  It's certainly possible that pygithub was quietly returning missing data into `None` where this will now error in those cases.  I'm not sure what to do besides just testing it against github.com (which I've done).

PR #20, opened 2019-07-03
OJ-4241: remove dangling pygithub dependency

eliwind:


eliwind:
Incorporated into PR 21

PR #24, opened 2019-07-09
Use new Jira users/search endpoint where possible

eliwind:


jf_agent/jira_download.py
```diff
@@ -273,30 +273,29 @@ def download_worklogs(jira_connection, issue_ids):
...
-        # a wildcard
-        users = _search_users(
-            jira_connection, gdpr_active, query='%', start_at=start_at, include_inactive=True
-        ) or _search_users(
-            jira_connection, gdpr_active, query='@', start_at=start_at, include_inactive=True
-        )
+        users = None
+
+        # different jira versions / different times, the way to list all users has changed. Try a few.
+        for q in [None, '', '%', '@']:
```
glenn-barnett: should this `for` loop be outside the `while True`?  so it paginates through `q=None`, then through `q=''`, then `q='%'` etc?

jf_agent/jira_download.py
```diff
@@ -273,30 +273,29 @@ def download_worklogs(jira_connection, issue_ids):
...
+def _jira_user_key(user_dict):
+    return user_dict.get('key', user_dict['accountId'])
 
 
 def _search_all_users(jira_connection, gdpr_active):
     jira_users = {}
     start_at = 0
+
+    # iterate through pages of results
     while True:
```
jnicoleoliveira: Do you need this while true?

jnicoleoliveira:
I approved before noticing what @glenn-barnett pointed out. I think that while True should be removed.

jf_agent/jira_download.py
eliwind: I thought the code was simpler this way, but you're right it's a little wasteful to be trying all the query strings on each page iteration.

PR #25, opened 2019-08-22
Repackaging etc

mattklein:


Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM python:3.7 AS py-deps
+COPY ./Pipfile ./Pipfile.lock ./
+RUN pip install pipenv && \
+    pipenv install --deploy --system --ignore-pipfile --clear
+
+FROM python:3.7-alpine
```
ametzger: nit: minor concern with alpine/musl vs debian/glibc performance, but I think we can wait to see how this performs

jf_agent/main.py
```diff
@@ -46,14 +46,43 @@
     download_issues,
     download_worklogs,
 )
+from jf_agent.session import retry_session
+
+JELLYFISH_API_BASE = 'https://jellyfish.co'
+# JELLYFISH_API_BASE = 'https://cf758842.ngrok.io'
```
ametzger: Probably want to remove dev URL here

jf_agent/main.py
```diff
@@ -86,106 +127,214 @@ def main():
...
+        print(f'ERROR: JELLYFISH_API_TOKEN not found in the environment.')
+        return
+
+    resp = requests.get(f'{JELLYFISH_API_BASE}/agent/config?api_token={api_token}')
+    if not resp.ok:
+        print(
+            f"ERROR: Couldn't get agent config info from {JELLYFISH_API_BASE}/agent/config "
+            f'using provided JELLYFISH_API_TOKEN (HTTP {resp.status_code})'
+        )
+        return
```
ametzger: nit: I think it would be better to use `sys.exit(1)` here (and other places where we `return` based on bad configuration). In cases where the agent is being run non-interactively (e.g. systemd timer unit, cron job) we would want to signal that something has gone wrong. Specifically in the API failure case (which could be transient) we would want to make sure we weren't silently failing.

jf_agent/main.py
```diff
@@ -355,18 +532,61 @@ def get_git_client(provider, git_url, skip_ssl_verification):
...
 
-def write_file(outdir, name, results):
+def write_file(outdir, name, results, compress):
     if isinstance(results, GeneratorType):
         results = list(results)
 
-    with open(f'{outdir}/{name}.json', 'w') as outfile:
-        json.dump(results, outfile, indent=2, default=str)
+    if compress:
+        with gzip.open(f'{outdir}/{name}.json.gz', 'wb') as outfile:
```
ametzger: Nice 🎉

jf_agent/main.py
mattklein: Yeah agreed we should do more to support meaningful exit codes. Previously it had been existing with status 1 in some but not all error scenarios. I started down a path of standardizing, but punted on it realizing that there were nuances that weren't necessarily clear calls (e.g., partial successes). I'll ticket out taking a harder look at this.

PR #31, opened 2019-09-27
Gather additional system-related diagnostics: memory, CPU, disk

mattklein:


jf_agent/main.py
```diff
@@ -208,14 +209,21 @@ def main():
...
 
-
     if run_mode == 'send_only':
         # Importantly, don't overwrite the already-existing diagnostics file
         pass
 
     else:
         diagnostics.open_file(outdir)
 
+        sys_diag_done_event = threading.Event()
```
ametzger: TIL

PR #36, opened 2019-11-08
[OJ-5239]: Support for custom field options

jeffbrainerd:
Should be pretty straight-forward. 

Tested for orthogonal-networks. 
Not sure if I can test our actual agent customers or if we simply don't have the creds.

### DIFF
--- a/jf_agent/jira_download.py
+++ b/jf_agent/jira_download.py
@@ -310,8 +310,10 @@ def download_customfieldoptions(jira_connection):
     opt_dict = {}
     for project in meta['projects']:
         for issue_type in project['issuetypes']:
-            for field in issue_type['fields'].values():
-                field_key = field['key']
+            for field_key, field in issue_type['fields'].items():
+                # we've seen variance in the location of the key field -- use it if it's available (OJ-5279)
+                if 'key' in field:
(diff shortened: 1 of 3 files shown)


jf_agent/jira_download.py
```diff
@@ -292,6 +292,48 @@ def download_worklogs(jira_connection, issue_ids):
...
 
+# Returns an array of CustomFieldOption items
+@diagnostics.capture_timing()
+@agent_logging.log_entry_exit(logger)
+def download_customfieldoptions(jira_connection):
+    print('downloading jira custom field options... ', end='', flush=True)
+    try:
+        meta = jira_connection.createmeta(projectIds=None, expand='projects.issuetypes.fields')
+    except JIRAError:
+        agent_logging.log_and_print(
```
mattklein: Should probably just be `print` to be consistent with other error handlers.

We're careful printing things to the logger in the agent since we want to make sure we don't print anything that could contain customer-sensitive info -- e.g., names of their things. This message that you have doesn't do that, but probably best to stay consistent with the other code anyway.

jeffbrainerd: :+1:

PR #37, opened 2020-01-08
[OJ-5628] S3 bucket name validation

ametzger:


jf_agent/main.py
```diff
@@ -403,6 +401,14 @@ def obtain_jellyfish_endpoint_info(config, creds):
     aws_secret_access_key = agent_config.get('aws_secret_access_key')
     git_instance_info = agent_config.get('git_instance_info')
 
+    # Validate the S3 prefix, bail out if invalid value is found.
+    if not re.fullmatch(r'^s3:\/\/[A-Za-z0-9:\/\-_]+\/[A-Za-z0-9:\/\-_]+$', s3_uri_prefix or ''):
+        print(
+            f"ERROR: The S3 bucket information provided by the API is invalid -- please contact Jellyfish",
+            file=sys.stderr,
```
mattklein: Probably best to keep consistent with the other error messages, which are (for better or worse) currently written to stdout, not stderr.

And maybe, in a similar vein, call it "the agent config endpoint" rather than "the API".

jf_agent/main.py
```diff
@@ -1,52 +1,50 @@
...
     download_projects_and_versions,
-    download_boards_and_sprints,
-    download_issue_batch,
+    download_resolutions,
+    download_users,
     download_worklogs,
-    download_customfieldoptions,
 )
 from jf_agent.session import retry_session
+from jira import JIRA
```
mattklein: I wonder why isort doesn't realize jira and stashy are third-party libs like yaml and sort them above the custom libs?

ametzger: FWIW, this appears to be a shortcoming of my `isort` plugin - it wasn't picking up the virtualenv from `pipenv`. I ran it manually specifying the virtualenv and everything worked:
`isort --virtual-env "$(pipenv --venv)" jf_agent/main.py`

PR #38, opened 2020-01-13
OJ 5636 - Handling Agent Failure when files fail to upload to s3

tina-n219:
**ToDo**:

- [x] Test! 
- [x] Get the stack trace to show up for the error
- [x] try writing to JSON instead of .txt for further expansion of this feature later on

**Purpose**: To notify the production channel when files don't all import into an S3 bucket during load and dump.

**Scope of this PR:** creating a `<service>_load_and_dump_exception.txt` file when there is an issue with loading into S3

**Output when an exception is raised**:
Screen Shot 2020-01-16 at 1 06 27 PM 1

### DIFF
--- a/jf_agent/__init__.py
+++ b/jf_agent/__init__.py
@@ -40,13 +40,17 @@ def pull_since_date_for_repo(server_git_instance_info, org_login, repo_id, commi
         return instance_pull_from_dt
 
 
-def write_file(outdir, filename_prefix, compress, results):
+def write_file(outdir, filename_prefix, compress, results, append=False):
     if compress:
         with gzip.open(f'{outdir}/{filename_prefix}.json.gz', 'wb') as outfile:
             outfile.write(json.dumps(results, indent=2, default=str).encode('utf-8'))
     else:
-        with open(f'{outdir}/{filename_prefix}.json', 'w') as outfile:
-            outfile.write(json.dumps(results, indent=2, default=str))
+        if append:
+            with open(f'{outdir}/{filename_prefix}.json', 'a') as outfile:
+                outfile.write(json.dumps(results, indent=2, default=str))
+        else:
+            with open(f'{outdir}/{filename_prefix}.json', 'w') as outfile:
+                outfile.write(json.dumps(results, indent=2, default=str))   
 
 
 class StrDefaultEncoder(json.JSONEncoder):
--- a/jf_agent/__init__.py
+++ b/jf_agent/__init__.py
@@ -48,15 +48,6 @@ def write_file(outdir, filename_prefix, compress, results):
         with open(f'{outdir}/{filename_prefix}.json', 'w') as outfile:
             outfile.write(json.dumps(results, indent=2, default=str))
 
-def write_status_file(outdir, type, status):
-    with open(f'{outdir}/status.json', 'r') as status_file:
-        status_json = json.load(status_file)
-        body = {
-            'type': type,
-            'status': status
-        }
-        status_json.append(body)
-        write_file(outdir, 'status', False, status_json)
 
 class StrDefaultEncoder(json.JSONEncoder):
     def __init__(self, *args, **kwargs):
--- a/jf_agent/__init__.py
+++ b/jf_agent/__init__.py
@@ -40,17 +40,13 @@ def pull_since_date_for_repo(server_git_instance_info, org_login, repo_id, commi
         return instance_pull_from_dt
 
 
-def write_file(outdir, filename_prefix, compress, results, append=False):
+def write_file(outdir, filename_prefix, compress, results):
     if compress:
         with gzip.open(f'{outdir}/{filename_prefix}.json.gz', 'wb') as outfile:
             outfile.write(json.dumps(results, indent=2, default=str).encode('utf-8'))
     else:
-        if append:
-            with open(f'{outdir}/{filename_prefix}.json', 'a') as outfile:
-                outfile.write(json.dumps(results, indent=2, default=str))
-        else:
-            with open(f'{outdir}/{filename_prefix}.json', 'w') as outfile:
-                outfile.write(json.dumps(results, indent=2, default=str))   
+        with open(f'{outdir}/{filename_prefix}.json', 'w') as outfile:
+            outfile.write(json.dumps(results, indent=2, default=str))   
 
 
 class StrDefaultEncoder(json.JSONEncoder):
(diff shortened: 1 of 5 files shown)


jf_agent/main.py
```diff
@@ -592,21 +594,28 @@ def get_basic_jira_connection(config, creds):
...
+            get_gh_projects(git_conn, config.git_include_projects, config.git_redact_names_and_urls),
+        )
 
-    api_repos = None
+        api_repos = None
+    except Exception as e:
+        write_file_txt(
+            config.outdir,
+            'git_load_and_dump_exception',
+            e,
```
eliwind: 1) This looks like it'll try to write out an `Exception` object as though it were text.  Is something implicitly string-ifying the exception in there somewhere?   
2) Text seems like the wrong format, perhaps? I'm imagining we may, over time, want more things in there (like, a boolean with an explicit "succeed/fail" rather than an implicit absence-of-a-thing meaning success; maybe a stack trace, maybe a string message, maybe other context over time), which makes me think of a structured `result.json` or some such thing

tina-n219: 1. yes - should be doing `repr(e)` to preserve the error along with its message
2. good point!

mattklein:
Meta-comment is that we might get more complete coverage of error conditions by looking for all of the locations where we call log_and_print with a level of logging.ERROR.  All of those occurrences are probably worthy of us putting an indication in the diagnostics file that an error was encountered and what the implication is (e.g., Jira data download incomplete, Git data download incomplete, S3 upload incomplete, etc.).

Perhaps what we should do is keep track of the highest severity message encountered (INFO, WARN, ERR) during each run.  Then upon completion, write an entry to the diagnostics file with the highest level encountered, along with separate indications as to whether Jira data is complete and Git data is complete.

The goal is for us to be able to support a couple of different conditions (at least) server-side:
 - If Jira/Git data isn't complete in the uploaded fileset, we handle that appropriately (e.g., we know not to blow away data we've got stored because the agent didn't upload it)
 - If agent encountered an error, that we're able to notify the customer about that in case they don't realize it

jf_agent/main.py
```diff
@@ -759,9 +773,14 @@ def get_git_client(config, creds):
                 session=retry_session(),
             )
         except Exception as e:
-            agent_logging.log_and_print(
-                logger, logging.ERROR, f'Failed to connect to Bitbucket:\n{e}', exc_info=True
-            )
+            body = {
+                'type': 'ERROR: Git - Failed to connect to Bitbuck',
```
jeffbrainerd: minor: bitbucket mispelled

jf_agent/__init__.py
```diff
@@ -46,7 +46,7 @@ def write_file(outdir, filename_prefix, compress, results):
             outfile.write(json.dumps(results, indent=2, default=str).encode('utf-8'))
     else:
         with open(f'{outdir}/{filename_prefix}.json', 'w') as outfile:
-            outfile.write(json.dumps(results, indent=2, default=str))
+            outfile.write(json.dumps(results, indent=2, default=str))
```
jeffbrainerd: spurious diff? I think there's a setting in PyCharm to remove trailing whitespace from each line

jf_agent/main.py
```diff
@@ -439,23 +441,48 @@ def print_all_jira_fields(config, jira_connection):
...
             )
 
+            with open(f'{config.outdir}/status.json', 'r') as status:
+                status_json = json.load(status)
+                body = {
+                    'type': 'Git',
+                    'status': 'failed'
+                }
+                status_json.append(body)
+                write_file(config.outdir, 'status', config.compress_output_files, status_json)
```
jeffbrainerd: the 3 blocks here are almost identical -- maybe create a helper function that takes in parameters, or else assign local vars and then execute the code once at the end of the function

jf_agent/__init__.py
```diff
@@ -48,6 +48,15 @@ def write_file(outdir, filename_prefix, compress, results):
...
             outfile.write(json.dumps(results, indent=2, default=str))
 
+def write_status_file(outdir, type, status):
+    with open(f'{outdir}/status.json', 'r') as status_file:
+        status_json = json.load(status_file)
+        body = {
+            'type': type,
+            'status': status
+        }
+        status_json.append(body)
```
jeffbrainerd: just want to make sure we've tested the cases where you are appending to an empty file and also appending to a non-empty file

jf_agent/main.py
```diff
@@ -47,7 +49,7 @@
 
 logger = logging.getLogger(__name__)
 
-JELLYFISH_API_BASE = 'https://jellyfish.co'
+JELLYFISH_API_BASE = 'http://localhost:8000'
```
jeffbrainerd: I think you don't want to check this in 😁

jf_agent/__init__.py
```diff
@@ -48,6 +48,15 @@ def write_file(outdir, filename_prefix, compress, results):
...
+        write_file(outdir, 'status', False, status_json)
```
seancrater: What happens when the status file already exists from a prior agent run? We end up with multiple statuses for the same type?

jf_agent/main.py
```diff
@@ -439,23 +441,28 @@ def print_all_jira_fields(config, jira_connection):
 @diagnostics.capture_timing()
 @agent_logging.log_entry_exit(logger)
 def download_data(config, endpoint_git_instance_info, jira_connection, git_connection):
+    write_file(config.outdir, 'status', config.compress_output_files, [])
```
jeffbrainerd: what's the purpose of this?

tina-n219: @jeffbrainerd This creates the `status.json` with just `[]` when download data is run. This is so that all the `write_files()` that get called after this add their results into the bracket - creating a properly formatted json. I wasn't sure which `write_files()` function calls would get called and in what combination with regards to when the `[]` should first initially be added. This lines avoids having to deal with that and just creates the file initially with the `[]`.

does this make sense?

jf_agent/__init__.py
tina-n219: Whenever the agent is run, a new directory with the timestamp is created and that is where it dumps the files it writes to. I'm not sure it will end up writing to an already existing status file - but I may be wrong.

jf_agent/__init__.py
tina-n219: So the first `write_file()` call I added that puts in a `[]` to the status file makes it such that these append calls are never on an empty file.

jf_agent/main.py
tina-n219: 🤦‍♀️ whoops

PR #44, opened 2020-03-06
OJ 6166 - agent failing on Bitbucket data

tina-n219:
Fix for failing agent that was throwing back

`  File "/usr/local/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

### DIFF
--- a/jf_agent/bb_download.py
+++ b/jf_agent/bb_download.py
@@ -21,9 +21,9 @@ def datetime_from_bitbucket_server_timestamp(bb_server_timestamp_str):
 
 def _normalize_user(user):
     return {
-        'id': user['id'],
-        'login': user['name'],
-        'name': user['displayName'],
+        'id': user.get('id', ''),
+        'login': user.get('name', ''),
+        'name': user.get('displayName', ''),
         'email': user.get('emailAddress', ''),
     }
 


jf_agent/bb_download.py
```diff
@@ -21,9 +21,9 @@ def datetime_from_bitbucket_server_timestamp(bb_server_timestamp_str):
 
 def _normalize_user(user):
     return {
-        'id': user['id'],
-        'login': user['name'],
-        'name': user['displayName'],
+        'id': user.get('id', ''),
+        'login': user.get('name', ''),
+        'name': user.get('displayName', ''),
```
jnicoleoliveira: Is there a case where all three are blank? I think if so, we could end up with a blank user when importing.
```
[ git_import.py ]
 with load_json(input_dir, 'users') as git_users:
        for user in tqdm(git_users, desc='loading users'):
            JFGithubUser.objects.update_or_create(
                instance=git_instance,
                user_id=user.get('id', f"{user['name']}<{user['email']}>"),
                defaults={'name': user['name'], 'login': user['login'], 'email': user['email']},
            )
```

PR #51, opened 2020-03-19
OJ-6251: Allow "no git configuration" case

jnicoleoliveira:


jf_agent/main.py
```diff
@@ -247,8 +247,8 @@ def obtain_config(args):
     jira_exclude_project_categories = set(jira_config.get('exclude_project_categories', []))
     jira_issue_jql = jira_config.get('issue_jql', '')
 
-    git_config = yaml_config.get('git', yaml_config.get('bitbucket', {}))
-    git_provider = git_config.get('provider', 'bitbucket_server')
+    git_config = yaml_config.get('git', {})
+    git_provider = git_config.get('provider', None)
     git_url = git_config.get('url', None)
```
eliwind: These defaults were in there to be backwards-compatible with old YML format.  It's possible that parallel etc are still using that old YML config file, so we may want to leave the defaults in.

jf_agent/main.py
```diff
@@ -247,8 +247,8 @@ def obtain_config(args):
     jira_exclude_project_categories = set(jira_config.get('exclude_project_categories', []))
     jira_issue_jql = jira_config.get('issue_jql', '')
 
-    git_config = yaml_config.get('git', yaml_config.get('bitbucket', {}))
```
ametzger: Does this mean that existing config files using the `bitbucket` section will stop working?

jnicoleoliveira: ah, that's a good point. i should make this backwards compatible then.

jf_agent/main.py
```diff
@@ -247,7 +247,7 @@ def obtain_config(args):
...
+    git_config = yaml_config.get('git', {})
     git_provider = git_config.get('provider', 'bitbucket_server')
     git_url = git_config.get('url', None)
```
ametzger: I believe this still isn't compatible with old yaml, for example (per the [old `example.yml`](https://github.com/Jellyfish-AI/jf_agent/blob/c1a7cf049d0e798594964db56fc884ecbe8f74b6/example.yml#L50-L52):
``` yaml
bitbucket:
  url: https://some-bitbucket-server.example.com
```

this code would discard that and instead disable git pulling. I think the right thing to do would be:

``` python
legacy_config = yaml_config.get('bitbucket')
if 'url' in legacy_config:
    git_url = legacy_config['url']
    git_provider = 'bitbucket_server'
else:
    git_config = yaml_config.get('git', {})
    git_provider = git_config.get('provider', 'bitbucket_server')
    git_url = git_config.get('url', None)
    # ...
```

jnicoleoliveira: This should be fixed now. I misunderstood the legacy formatting.

PR #54, opened 2020-04-03
OJ-6415: Add "log_as_exception option" to Gitlab log and print

jnicoleoliveira:


jf_agent/git/gitlab_adapter.py
```diff
@@ -179,8 +179,6 @@ def get_pull_requests(
     ) -> List[NormalizedPullRequest]:
         print('downloading gitlab prs... ', end='', flush=True)
 
-        normalized_prs = []
-
```
eliwind, line 183: good call

PR #55, opened 2020-04-08
OJ 4655

tina-n219:


jf_agent/main.py
```diff
@@ -211,7 +210,7 @@ class BadConfigException(Exception):
 
 JellyfishEndpointInfo = namedtuple(
     'JellyfishEndpointInfo',
-    ['s3_uri_prefix', 'aws_access_key_id', 'aws_secret_access_key', 'git_instance_info'],
+    ['git_instance_info'],
```
eliwind, line 213: it seems a little bizarre to have a namedtuple with only one thing in it

jf_agent/main.py
```diff
@@ -444,35 +443,16 @@ def obtain_jellyfish_endpoint_info(config, creds):
         raise BadConfigException()
```
eliwind, line 443: absolete logging?

jf_agent/main.py
tina-n219: Yea, I should update this - doesn't seem to serve a purpose anymore.

tina-n219:
> How does it change things to be uploading files one by one instead of s3 sync?
> 
> One difference is that s3 sync does all the files in parallel, so I expect this to take longer. I'm guessing there are differnt failure modes too, where maybe this way we can more easily wind up in a state with some files and not others?

@eliwind Files will take different lengths of time to upload, and yes, failures midway through will result in a partial upload. Just like there is a status for jira download & git download, there may need to be some loud failure that notifies slack if the upload is impartial

PR #56, opened 2020-04-08
OJ 5718 - agent versioning 

tina-n219:


jf_agent/diagnostics.py
```diff
@@ -21,6 +21,16 @@ def _write_diagnostic(obj):
     DIAGNOSTICS_FILE.flush()
 
 
+def capture_agent_version():
+    git_head_hash = subprocess.check_output(['git', 'rev-parse', 'HEAD']).strip().decode('utf-8')
```
ametzger, line 25: we strip the git repository out when we do a `docker build` of this repo - let's chat about what the right approach is here

PR #57, opened 2020-04-14
Oj-6521: bitbucket cloud support & more jira configuration

eliwind:


Pipfile
```diff
@@ -15,5 +15,6 @@ dateparser = "*"
 jsonstreams = "*"
 psutil = "*"
 python-gitlab = "*"
+numpy = ">= 1.18, < 1.19"
```
ametzger, line 18: this doesn't seem to be used anywhere?

mattklein: It's this line in `jira_download.py`:
```
    issue_ids_for_threads = numpy.array_split(issue_ids_to_download, num_threads_to_use)
```

Could write it without using numpy, but I thought it was better to let them work out the corner cases. Adding numpy increased the image size from 312 MB to 393 MB, which seems OK.

ametzger: the code to do that seems straightforward enough I'm not convinced it's worth 79M:
``` python
def chunk(l, n):
    for i in range(0, len(l), n):
        yield l[i:i+n]
```

mattklein: OK fair enough, I'll add that as a TODO

PR #61, opened 2020-04-17
OJ 4655 

tina-n219:


jf_agent/main.py
```diff
@@ -136,21 +137,19 @@ def main():
...
-            jellyfish_endpoint_info.aws_secret_access_key,
+            config,
+            creds
         )
     else:
         print(f'\nSkipping send_data because run_mode is "{config.run_mode}"')
         print(f'You can now inspect the downloaded data in {config.outdir}')
         print(f'To send this data to Jellyfish, use "-m send_only -od {config.outdir}"')
 
+    write_file(config.outdir, 'status', config.compress_output_files, status_json)
```
eliwind, line 152: Don't we want to send this to the backend?  Should that happen befire `send_data`?

jf_agent/main.py
```diff
@@ -537,22 +514,25 @@ def download_data(
...
+        headers = {'Jellyfish-API-Token': creds.jellyfish_api_token}
+        r = requests.get(f'{base_url}/endpoints/agent/signed-url?filename={filename}&timestamp={timestamp}', headers=headers).json()
+        signed_url = r["signedUrl"]
+        path_to_obj = r['objectPath']
+        return signed_url, path_to_obj
+    
+    def upload_file(filename, signed_url, path_to_obj):
+        with open(f'{config.outdir}/{filename}', 'rb') as f:
+            # If successful, returns HTTP status code 204
+            upload_resp = requests.post(signed_url['url'], data=signed_url['fields'], files={'file': (path_to_obj, f)})
```
eliwind, line 531: I think we want retries around this; S3 is flaky.  Since we're not using the boto library, we'll want something like `jf_agent.session.retry_session` here.

jf_agent/main.py
```diff
@@ -537,22 +514,25 @@ def download_data(
...
+            print(f'File {filename} upload HTTP status code: {upload_resp.status_code}')
 
     # Compress any not yet compressed files before sending
-    for fname in glob(f'{outdir}/*.json'):
+    for fname in glob(f'{config.outdir}/*.json'):
```
eliwind, line 535: Should we mimic `s3 sync` and do these in parallel?  The files can be large, seems like we'll be primarily gated by network, so having them each go in their own thread could be useful.

jf_agent/main.py
```diff
@@ -561,19 +550,46 @@ def _s3_cmd(cmd):
...
 
-    output_basedir, output_dir_timestamp = os.path.split(outdir)
-    s3_uri_prefix_with_timestamp = os.path.join(s3_uri_prefix, output_dir_timestamp)
-    done_file_path = f'{os.path.join(outdir, ".done")}'
+    signed_urls = get_signed_url(os.listdir(config.outdir))
+
+    threads = []
+    thread_exceptions = [None] * (len(signed_urls) + 1)
+    for index, file_info in list(enumerate(signed_urls)):
+        # file_info shape {filename: (path_to_obj, signed_url object)}
```
eliwind, line 558: See PR on the JF side, but I think the shape of the data here doesn't make sense.  Why would we have a list of single-element dicts?  If this were simply a dict of filename to URL, this whole loop could be something like 

```
threads = [
    threading.Thread(target=upload_file, args=[index, filename, url])
    for index, (filename, url) in enumerate(signed_urls.items())
]

for t in threads:
    t.start()
```

jf_agent/main.py
```diff
@@ -136,21 +138,20 @@ def main():
...
+            config,
+            creds
         )
+        status_json.append(s3_upload_status)
     else:
         print(f'\nSkipping send_data because run_mode is "{config.run_mode}"')
         print(f'You can now inspect the downloaded data in {config.outdir}')
         print(f'To send this data to Jellyfish, use "-m send_only -od {config.outdir}"')
 
+    write_file(config.outdir, 'status', config.compress_output_files, status_json)
```
eliwind, line 154: is anything ever sending the status json?

tina-n219: 🤦‍♀️

jf_agent/main.py
```diff
@@ -561,19 +550,46 @@ def _s3_cmd(cmd):
...
+    
+    print(f'Joining {len(threads)} threads')
+
+    for process in threads:
+        process.join()
+
+    # if all threads failed
+    if any(thread_exceptions):
+        status = {'type': 'S3 Upload', 'status': 'failed'}
+    else:
```
eliwind, line 579: Perhaps raise the first one, if there's at least one exception?   And/or log them here?  I think as-is, if a thread fails, we'll never find out the details of what happened.

jf_agent/main.py
```diff
@@ -567,19 +552,40 @@ def _s3_cmd(cmd):
...
+
+    if any(thread_exceptions):
+        print(f'ERROR: not all files uploaded to S3. Files have been saved locally. Once connectivity issues are resolved, try running the Agent in send_only mode.')
+        return
+
+    # creating .done file
+    done_file_path = f'{os.path.join(config.outdir, ".done")}'
     if os.path.exists(done_file_path):
         print(
             f'ERROR: {done_file_path} already exists -- has this data already been sent to Jellyfish?'
```
eliwind, line 583: I'm not sure this check is helpful.

In the case that the S3 upload of the .done file fails, you're left with a .done file sitting on the disk with the other generated files; if you try to send again, we'd hit this error, which seems unhelpful.  Why not just let them re-send?

jf_agent/git/__init__.py
```diff
@@ -1,3 +1,4 @@
+import os
```
eliwind, line 1: extra import still :)

jf_agent/jf_jira/__init__.py
```diff
@@ -1,6 +1,7 @@
 from datetime import datetime
 from itertools import chain
 import logging
+import os
```
eliwind, line 4: extra import?

PR #64, opened 2020-04-27
OJ-6323

aregna:


jf_agent/jf_jira/jira_download.py
```diff
@@ -713,3 +713,200 @@ def _search_users(
...
+                    missing_repositories[repo_name] = {
+                        'name': repo_name,
+                        'url': repo_url,
+                        'instance_type': instance_type,
+                    }
+
+    print()
+
+    try:
+        _remove_mismatched_repos(
```
jnicoleoliveira, line 756: I know what this does, but a comment here about the what the side effect of this function is might be useful?

jf_agent/jf_jira/jira_download.py
```diff
@@ -713,3 +713,200 @@ def _search_users(
...
+        )
+
+    elif config.git_provider == 'gitlab':
+
+        from jf_agent.git.gitlab_adapter import GitLabAdapter
+
+        gl_adapter = GitLabAdapter(git_connection)
+
+        projects = gl_adapter.get_projects(config.git_include_projects, False)
+        repos = gl_adapter.get_repos(
```
jnicoleoliveira, line 817: nit: i think you could return early for these repos. this may just be a personal reference, in which case feel free to ignore me. 
```
...
if =='':
  return gl_adapeter.get_repos()
if== ""
  return other_way_to_get_repos()
raise ValueError()
```

jf_agent/jf_jira/jira_download.py
```diff
@@ -713,3 +713,200 @@ def _search_users(
...
+
+        repo_name = repo['name']
+        repo_url = repo['url']
+
+        # Jira sometimes responds with a different version of names/urls than
+        # what is returned from their respective git APIs. In order to cross reference
+        # the git repos given to us, we will try to query with something more similar:
+        # (url ex.) gitlab.com/group/@group/@subgroup/@myrepo >> gitlab.com/group/subgroup/myrepo
+        # (name ex.) @group@subgroup@myrepo >> myrepo
+        if '@' in repo_url:
```
jnicoleoliveira, line 892: fun times

jf_agent/main.py
```diff
@@ -15,14 +15,27 @@
...
 logger = logging.getLogger(__name__)
 
 JELLYFISH_API_BASE = 'https://jellyfish.co'
-VALID_RUN_MODES = ('download_and_send', 'download_only', 'send_only', 'print_all_jira_fields')
+VALID_RUN_MODES = (
+    'download_and_send',
+    'download_only',
+    'send_only',
+    'print_all_jira_fields',
+    'scan_jira_for_missing_repos',
```
mattklein, line 37: This run mode is in the same family as "print_all_jira_fields". It might be more understandable to the user if we called it something more parallel to that, something like "print_apparently_missing_git_repos".

mattklein: Also note the README.md should be updated for this new run mode

jf_agent/jf_jira/jira_download.py
```diff
@@ -713,3 +713,200 @@ def _search_users(
...
+            f'WARNING: Got an error when trying to cross reference repos discoverd by '
+            f'Jira with Git repos: {e}\nSkipping this process and returning all repos '
+            f'discovered by Jira...'
+        )
+
+    result = [{'name': k, 'url': v['url']} for k, v in missing_repositories.items()]
+    return result
+
+
+def _get_repos_from_git(git_connection, config):
```
mattklein, line 770: I wonder if we can structure this so that the code in `jira_download.py` remains free of having to do anything with a git_connection, and that Git-specific code stays isolated in the modules inside of the git package. This would help prevent us from, e.g., adding a new Git provider down the road and forgetting that the code in _get_repos_from_git would need to be updated for it.

jf_agent/jf_jira/__init__.py
```diff
@@ -61,6 +62,18 @@ def print_all_jira_fields(config, jira_connection):
         print(f"{f['key']:30}\t{f['name']}")
 
 
+@diagnostics.capture_timing()
+@agent_logging.log_entry_exit(logger)
+def print_missing_repos_found_by_jira(issues_to_scan, config, jira_connection, git_connection):
+    missing_repos = download_missing_repos_found_by_jira(
+        issues_to_scan, config, jira_connection, git_connection
+    )
+    print(f'\nJira found {len(missing_repos)} repos potentially missing from Jellyfish')
```
mattklein, line 71: Possible alternate suggestions:

 - Scanning the "Development" field on the Jira issues revealed XX Git repos apparently missing from Jellyfish

 - Jira issues refer to XX Git repos that are missing from Jellyfish

jf_agent/main.py
```diff
@@ -64,6 +77,15 @@ def main():
             "http://172.17.0.1:8000 (if running the agent container with --network bridge)"
         ),
     )
+    parser.add_argument(
+        '--for-print-missing-repos-issues-updated-within-last-x-months',
+        type=int,
+        choices=range(1, 6),
```
mattklein, line 83: Note that `list(range(1, 6))` is `[1, 2, 3, 4, 5]`, so I think you want `range(1, 7)`

jf_agent/main.py
```diff
@@ -63,6 +74,15 @@ def main():
             "http://172.17.0.1:8000 (if running the agent container with --network bridge)"
         ),
     )
+    parser.add_argument(
+        '--for-print-missing-repos-issues-updated-within-last-x-months',
```
tina-n219, line 78: So this is a really long name for a flag, thoughts on an abbreviation, like how `--output-basedir` has `-ob`? Or does it behave different as an argument that expects an `int` value afterwards? 🤔

jf_agent/main.py
```diff
@@ -590,6 +632,33 @@ def upload_file(filename, path_to_obj, signed_url):
...
+
+    print(f'Fetching Jira issues that are missing Git repo data in Jellyfish...')
+
+    resp = requests.get(
+        f'{base_url}/endpoints/agent/unlinked-dev-issues',
+        headers={'Jellyfish-API-Token': api_token},
+        params=params,
+    )
+
+    data = resp.json()
```
tina-n219, line 651: want to ensure the resp is `ok` before parsing?

PR #69, opened 2020-05-06
OJ-6553: Migrate api_base to app.jellyfish.co

jnicoleoliveira:


jnicoleoliveira:
> Will our new scheme break older agents, that don't yet pull down this new code? Or will we have redirection in place at jellyfish.co?

Old versions of the agent that use the `jellyfish.co` domain will be backwards compatible after this change.

PR #75, opened 2020-05-26
OJ-6983: harden github rate limiting

eliwind:


jf_agent/git/github_client.py
```diff
@@ -87,33 +87,52 @@ def get_json(self, url):
...
+                        logging.ERROR,
+                        f'Request to {url} has failed {i} times -- giving up!',
                     )
-                    time.sleep(reset_wait.seconds)
-                    continue
+                    raise
+
+                # rate-limited!  Sleep until it's ok, then try again
+                reset_time = datetime.fromtimestamp(int(ratelimit_reset), pytz.utc)
+                now = datetime.utcnow().replace(tzinfo=pytz.utc)
```
jeffbrainerd, line 114: do you need the `.replace`? should be all utc right?

jf_agent/git/github_client.py
```diff
@@ -87,33 +87,52 @@ def get_json(self, url):
...
 
-                raise
+                reset_wait_in_seconds = reset_wait.total_seconds()
+
+                # Sometimes github gives a reset time in the
+                # past. In that case, wait for 5 mins just in case.
+                if reset_wait_in_seconds <= 0:
+                    reset_wait_in_seconds = 300
+
+                # Soemtimes github gives a reset time way in the
```
jeffbrainerd, line 124: super-minor: typo `sometimes`

jf_agent/git/github_client.py
eliwind: @jeffbrainerd no, it's needed.  `datetime.utcnow()` returns a datetime relative to UTC, but without timezone set.

PR #79, opened 2020-07-06
OJ-6718: Support Multiple Instances of Git

jnicoleoliveira:


jf_agent/main.py
```diff
@@ -605,7 +372,16 @@ def upload_file(filename, path_to_obj, signed_url):
...
     print('Sending data to Jellyfish... ')
 
-    signed_urls = get_signed_url(os.listdir(config.outdir))
+    # obtain file names from the directory
+    _, directories, filenames = next(os.walk(config.outdir))
+
+    for directory in directories:
+        path = os.path.join(config.outdir, directory)
+        for file_name in os.listdir(path):
+            print(f'{directory}/{file_name}')
```
jnicoleoliveira, line 381: remove this line

jf_agent/git/__init__.py
```diff
@@ -261,28 +264,56 @@ def get_git_client(config, creds):
...
+    outdir: str,
+    compress_output_files: bool,
+    git_connection,
+):
+
+    # use the unique git instance agent key to collate files
+    instance_slug = endpoint_git_instance_info['slug']
+    instance_key = endpoint_git_instance_info['key']
+    outdir = f'{outdir}/git_{instance_key}'
+    os.mkdir(outdir)
```
mattklein, line 279: Not sure why we need a slug and a key -- what's the difference?

Also, if we're creating filenames based on key, we should probably edit that the user doesn't make it ridiculously long, and that it only includes ascii chars.

mattklein, line 279: Wondering about the rationale for using subdirectories for each instance. Are you sure it's necessary/desirable? If we guarantee that the key is unique, should we just write all files into a single flat directory instead?

jf_agent/main.py
mattklein: What does this mean?

jf_agent/main.py
```diff
@@ -605,7 +372,16 @@ def upload_file(filename, path_to_obj, signed_url):
 
     print('Sending data to Jellyfish... ')
 
-    signed_urls = get_signed_url(os.listdir(config.outdir))
+    # obtain file names from the directory
+    _, directories, filenames = next(os.walk(config.outdir))
```
mattklein, line 376: This doesn't look right -- doesn't `next` just get a single item?

Though see my comment above -- maybe we should just write all the files into the same directory and keep the rest of this code more or less unchanged.

jf_agent/main.py
```diff
@@ -150,10 +138,9 @@ def main():
             if config.run_mode_includes_download:
                 download_data_status = download_data(
                     config,
+                    creds,
                     jellyfish_endpoint_info.jira_info,
-                    jellyfish_endpoint_info.git_instance_info,
-                    jira_connection,
-                    git_connection,
+                    jellyfish_endpoint_info.git_instance_info,  # todo
```
mattklein, line 143: What's the TODO here for? Has this been addressed?

jf_agent/main.py
```diff
@@ -528,28 +261,62 @@ def obtain_jellyfish_endpoint_info(config, creds):
...
+    for git_config in config.git_configs:
+        git_connection = get_git_client(
+            git_config, creds, skip_ssl_verification=config.skip_ssl_verification
+        )
+
+        if is_multi_git_config:
+            instance_slug = git_config.git_instance_slug
+            instance_info = endpoint_git_instances_info.get(instance_slug)
+        else:
+            # support legacy single-git support, of which assumes only one available git instance
```
mattklein, line 309: Type-o: "of which" -> "which"

jf_agent/main.py
```diff
@@ -528,28 +261,62 @@ def obtain_jellyfish_endpoint_info(config, creds):
...
+            instance_info = list(endpoint_git_instances_info.values())[0]
```
mattklein, line 310: This relies on a change first having been deployed on the Jellyfish side, right?

jf_agent/config_file_reader.py
```diff
@@ -0,0 +1,311 @@
...
+    # gitlab must be in whitelist mode
+    if git_provider == 'gitlab' and (git_exclude_projects or not git_include_projects):
+        print(
+            'ERROR: GitLab requires a list of projects (i.e., GitLab top-level groups) '
+            'to pull from. Make sure you set `include_projects` and not `exclude_projects`, and try again.'
+        )
+        raise BadConfigException()
+
+    # BBCloud must be in whitelist mode
+    if git_provider == 'bitbucket_cloud' and (git_exclude_projects or not git_include_projects):
```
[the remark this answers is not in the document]
jnicoleoliveira: I agree that seems correct, but this was directly ported over from the original code (see `main.py`). Not sure if thats a pre-existing bug, or if there was intent behind this.

jf_agent/config_file_reader.py
```diff
@@ -0,0 +1,311 @@
...
+        git_redact_names_and_urls=git_config.get('redact_names_and_urls', False),
+        gitlab_per_page_override=git_config.get('gitlab_per_page_override', None),
+        # legacy fields ===========
+        git_include_bbcloud_projects=list(git_include_bbcloud_projects),
+        git_exclude_bbcloud_projects=list(git_exclude_bbcloud_projects),
+    )
+
+
+class BadConfigException(Exception):
+    pass
```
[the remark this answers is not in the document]
jnicoleoliveira: Yup, you're right. Refactor error.

jf_agent/main.py
jnicoleoliveira: Yup, it's been addressed, it's just an artifact.

jf_agent/main.py
jnicoleoliveira: Nope. The shape returning from the api is the actually same, it just has more items. Grabbing the first (and only item) is what we are doing currently.

jf_agent/main.py
jnicoleoliveira: This is right. `os.walk` returns an iterator, with each item being the next level of subdirectories. (I only want the first)

If I used `listdir()`, it would not show me which were directories, and which were filenames.

jf_agent/main.py
jnicoleoliveira: I will add a comment to clarify this code. Since the code creates subdirectories, I want to make sure that all those paths are included in `get_signed_urls`, otherwise we will not include them in their upload to s3.

README.md
```diff
@@ -78,6 +78,17 @@ JIRA_PASSWORD=...
...
+#### Multiple Git Instances
+If using multiple git instances, all git credentials must be prefixed with the corresponding `creds_envvar_prefix` provided in the config.yml.
+
+For example, if the  `creds_envvar_prefix` is set to ORG1 for a bitbucket instance, the configuration would include the following variables:
+```
+# ... other variables
+ORG1_BITBUCKET_USERNAME=...
+ORG1_BITBUCKET_PASSWORD=...
+```
+When in multi-git mode, `creds_envar_prefix` is required for all git instances. See the example.yml for more details.
```
ametzger, line 90: typo: creds_env*v*ar_prefix

jf_agent/git/__init__.py
```diff
@@ -261,28 +264,55 @@ def get_git_client(config, creds):
...
+def load_and_dump_git(
+    config: GitConfig,
+    endpoint_git_instance_info: dict,
+    outdir: str,
+    compress_output_files: bool,
+    git_connection,
+):
+    # use the unique git instance agent key to collate files
+    instance_slug = endpoint_git_instance_info['slug']
+    instance_key = endpoint_git_instance_info['key']
```
ametzger, line 276: what is the `instance_key`?

jnicoleoliveira: This is the agent file key that gets used to generate a distinguishable file prefix.

PR #93, opened 2020-11-30
[OJ-8356] [OJ-8365] More diagnostics

ametzger:
- Add a new `validate` mode that will test basic Jira connectivity
- Add some additional github enterprise validation
- be louder if there are no jira users found


mattklein (commented):
Looks like we could also use an update to README.md for this new mode!

jeffbrainerd (approved):
This looks like a great step forward!

mattklein (approved):

PR #97, opened 2020-12-21
OJ-8678: Catch Gitlab AttributeError

jnicoleoliveira:


eliwind (commented):
I feel like it would be helpful in this case to log what is in the response, vs just skip it.  Every single pr had this... do we really want to just go “I guess there was no approval” for all of them?

eliwind (approved):

PR #100, opened 2021-01-12
OJ-8692: Upload agent build timestamp to S3, along with SHA

splichte:
Allows for more detailed messaging to be logged by the Jellyfish API.

### DIFF
--- a/jf_agent/diagnostics.py
+++ b/jf_agent/diagnostics.py
@@ -27,8 +27,7 @@ def _write_diagnostic(obj):
 def capture_agent_version():
     git_head_hash = os.getenv('SHA')
     build_timestamp = os.getenv('BUILDTIME')
-    _write_diagnostic({'type': 'agent_version', 'sha': git_head_hash})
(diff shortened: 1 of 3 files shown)


ametzger (approved):

jf_agent/diagnostics.py
```diff
@@ -26,7 +26,9 @@ def _write_diagnostic(obj):
 
 def capture_agent_version():
     git_head_hash = os.getenv('SHA')
+    build_timestamp = os.getenv('BUILDTIME')
     _write_diagnostic({'type': 'agent_version', 'sha': git_head_hash})
+    _write_diagnostic({'type': 'agent_version', 'timestamp': build_timestamp})
```
mattklein, line 31: It's probably a bit more consistent with the other diagnostics to include these into the same item, so instead do:
```
_write_diagnostic({'type': 'agent_version', 'sha': git_head_hash, 'timestamp': build_timestamp})
```

mattklein (approved):

PR #110, opened 2021-02-01
[OJ-9084]: Jira changelog histories can get incorrectly omitted

jeffbrainerd:
Some changelog history items do not have `fieldId`, so use 'field` if available. Also include some additional safeguards.

### DIFF
--- a/jf_agent/jf_jira/jira_download.py
+++ b/jf_agent/jf_jira/jira_download.py
@@ -495,10 +495,7 @@ def download_necessary_issues(
                     new_issues_this_batch.append(issue)
             prog_bar.update(len(new_issues_this_batch))
 
-            if not include_fields and not exclude_fields:
-                yield new_issues_this_batch
-            else:
-                yield _filter_changelogs(new_issues_this_batch, include_fields, exclude_fields)
+            yield _filter_changelogs(new_issues_this_batch, include_fields, exclude_fields)
 
     for t in threads:
         t.join()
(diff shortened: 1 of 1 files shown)


jf_agent/jf_jira/jira_download.py
```diff
@@ -495,7 +495,10 @@ def download_necessary_issues(
                     new_issues_this_batch.append(issue)
             prog_bar.update(len(new_issues_this_batch))
 
-            yield _filter_changelogs(new_issues_this_batch, include_fields, exclude_fields)
+            if not include_fields and not exclude_fields:
+                yield new_issues_this_batch
+            else:
```
eliwind, line 500: I think this is unnecessary — if there’s no include or exclude the other block already handles the case properly

jf_agent/jf_jira/jira_download.py
```diff
@@ -505,18 +508,22 @@ def download_necessary_issues(
...
+        # Skip items that are a change of a field that's filtered out
         for i in items:
-            if 'fieldId' not in i:
-                continue
-            if include_fields and i['fieldId'] not in include_fields:
+            field_id_field = _get_field_identifier(i)
+            if not field_id_field:
+                logger.warning(
+                    f"OJ-9084: Changlog history item with no 'fieldId' or 'field' key: {i}"
+                )
```
mattklein, line 517: Best practice in agent is to use `log_and_print` instead of `logger.warning`.

Also we probably shouldn't include `{i}` in the output, in case that contains customer-sensitive data. `{i.keys()}` should be ok though.

jeffbrainerd: :+1:

jf_agent/jf_jira/jira_download.py
```diff
@@ -530,7 +536,10 @@ def _strip_changelog(c):
         # copy a changelog, stripping excluded fields from the history
         return {**c, 'histories': list(_strip_changelog_histories(c['histories']))}
 
-    return [{**i, 'changelog': _strip_changelog(i['changelog'])} for i in issues]
+    if not include_fields and not exclude_fields:
+        return issues
+    else:
+        return [{**i, 'changelog': _strip_changelog(i['changelog'])} for i in issues]
```
jeffbrainerd, line 542: Correct @eliwind not strictly necessary but I think a small improvement in efficiency and unexpected data issues down the road...

mattklein (approved):

PR #124, opened 2021-04-27
[OJ-10195] Skip PRs that throw `ChunkedEncodingError` from Bitbucket Server

ametzger:
It seems that some PRs will throw a `ChunkedEncodingError` consistently when we try and fetch them from Bitbucket Server. This seems like a bug in Bitbucket Server's handling of binary files, so if we see it just print a debug error message and move on.

### DIFF
--- a/jf_agent/git/bitbucket_server.py
+++ b/jf_agent/git/bitbucket_server.py
@@ -3,7 +3,7 @@ import logging
 import stashy
 import pytz
 from tqdm import tqdm
-from requests.exceptions import RetryError
+from requests.exceptions import RetryError, ChunkedEncodingError
 from jf_agent.git import pull_since_date_for_repo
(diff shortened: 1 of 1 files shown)


alison-j (approved):
Looks good. What state will this leave us in? (That is, will we try this on each run? That'd probably be for the best; we'll get the data quickly if it was a transient error, and it's okay that we'll keep getting this each time we run in the current case.)

ametzger:
> Looks good. What state will this leave us in? (That is, will we try this on each run? That'd probably be for the best; we'll get the data quickly if it was a transient error, and it's okay that we'll keep getting this each time we run in the current case.)

I believe that we will throw this each time we hit it with a PR that is within the time window the API returns for pulling PRs, so I think it will only happen once. Is that alright? I'm assuming we don't want to continuously re-pull these

alison-j:
Update seems great.

PR #134, opened 2021-06-03
OJ-10065: Add config option to send agent config information to Jellyfish

daviskeene:
# Description
Include a new flag to `config.yml` called `send_agent_config` which will, if `True`, send the customers' config.yml file to the s3 bucket.

# Testing Method
Set flag to `True` in `config.yml`:

```
daviskeene@Davis-Macbook-Pro:~$ aws s3 ls s3://jellyfish-dev-agent-upload/orthogonal-networks/20210603_183816/
2021-06-03 14:38:50          0 .done
2021-06-03 14:38:50        658 config.yml
2021-06-03 14:38:49        478 diagnostics.json.gz
2021-06-03 14:38:49       7849 jf_agent.log
2021-06-03 14:38:49        135 status.json.gz
```

Flag not included in `config.yml`:
```
daviskeene@Davis-Macbook-Pro:~/jf_agent$ aws s3 ls s3://jellyfish-dev-agent-upload/orthogonal-networks/20210603_184116/
2021-06-03 14:42:48          0 .done
2021-06-03 14:42:47        499 diagnostics.json.gz
2021-06-03 14:42:47       7847 jf_agent.log
2021-06-03 14:42:47        135 status.json.gz
```

### DIFF
--- a/example.yml
+++ b/example.yml
@@ -6,7 +6,9 @@ global:
   # Set this to True to skip verification of server SSL certificates.  This might
   # be useful if your Jira / Bitbucket server doesn't have a valid SSL certificate.
   no_verify_ssl: False
-
+  # Set this to False if you would not like to send your config.yml file to Jellyfish.
+  # We use this as a diagnostic tool for customer assistance.
+  send_agent_config_in_payload: True
 
 #############################
 # 2. Jira configuration
--- a/example.yml
+++ b/example.yml
@@ -8,7 +8,7 @@ global:
   no_verify_ssl: False
   # Set this to True if you would like to send your config.yml file to Jellyfish.
   # We use this as a diagnostic tool for customer assistance.
-  send_agent_config: False
+  send_agent_config: True
 
 #############################
 # 2. Jira configuration
--- a/example.yml
+++ b/example.yml
@@ -6,9 +6,9 @@ global:
   # Set this to True to skip verification of server SSL certificates.  This might
   # be useful if your Jira / Bitbucket server doesn't have a valid SSL certificate.
   no_verify_ssl: False
-  # Set this to False if you would not like to send your config.yml file to Jellyfish.
+  # Set this to True if you would like to send your config.yml file to Jellyfish.
   # We use this as a diagnostic tool for customer assistance.
-  send_agent_config_in_payload: True
+  send_agent_config: False
 
 #############################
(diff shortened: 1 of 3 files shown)


example.yml
```diff
@@ -6,7 +6,9 @@ global:
   # Set this to True to skip verification of server SSL certificates.  This might
   # be useful if your Jira / Bitbucket server doesn't have a valid SSL certificate.
   no_verify_ssl: False
-
+  # Set this to False if you would not like to send your config.yml file to Jellyfish.
+  # We use this as a diagnostic tool for customer assistance.
+  send_agent_config_in_payload: True
```
ametzger, line 11: this should default to `False` here, we would want customers to toggle it on in specific diagnostic situations

ametzger (changes requested):
One small change, overall looks great

example.yml
daviskeene: Sure, reason I had it set to `True` was because of this comment in the ticket's description:

> Let's add a configuration option to the agent called "send_agent_config_in_payload", which can be set to True or False. Let's default it to True in the example config.yml and the agent's setup instructions, with a comment indicating that the customer can flip it to False if their config.yml contains any sensitive information (but that this is uncommon).

ametzger: Ah I hadn't read that. IMO this is safer, but @mattklein if you feel strongly about it I could be convinced otherwise

ametzger (approved):

example.yml
mattklein: I do -- reasoning is farther down in the Jira ticket

jf_agent/main.py
```diff
@@ -492,6 +494,11 @@ def upload_file(filename, path_to_obj, signed_url):
         )
         return
 
+    # If sending agent config flag is on, upload config.yml to s3 bucket
```
mattklein, line 497: There's reason why this can't be done as another file to be uploaded by `upload_file_from_thread` in the block around line 476, is there?

mattklein (commented):
Looks great -- a couple comments...

jf_agent/main.py
daviskeene: Now that you mention it, I probably could add config.yml to the `filenames` list before that line. This just seemed more explicit to me.

PR #135, opened 2021-06-08
OJ-10525: Include disk space and memory statistics in agent validation

daviskeene:
## Description
Currently, the agent’s “validate” mode checks that the agent can contact Jira, that it can enumerate Jira users, and that it has access to the Jira projects it’s configured for. This PR adds the following functionality:

- Show how much RAM is available in the context it’s running in

- Show how much disk space is available on the output directory (`/home/jf_agent/output`)

## Testing Strategy
Local agent "validate" mode runs
Agent output available upon request

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -146,17 +146,39 @@ def main():
         # Git diagnostics
         import jf_agent.git as git
 
-        print("Git details:")
         git_configs = config.git_configs
         for i, git_config in enumerate(git_configs, start=1):
-            print(f"Git provider: {git_config.git_provider} ({i}/{len(git_configs)})")
-            print(f"Included Projects: {git_config.git_include_projects}")
+            print(f"Git details for instance {i}/{len(git_configs)}:")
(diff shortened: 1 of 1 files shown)


jf_agent/main.py
```diff
@@ -143,6 +143,29 @@ def main():
...
 
+        # Memory & Disk Usage diagnostics
+        import os
+        import shutil
+
+        print("Memory & Disk Usage:")
+        # From here:
+        # https://stackoverflow.com/questions/51248144/getting-cpu-and-memory-usage-of-a-docker-container-from-within-the-dockerized-ap
+        used = int(
+            os.popen('cat /sys/fs/cgroup/memory/memory.usage_in_bytes').readlines()[0].strip()
```
jeffbrainerd, line 154: wondering if we want to wrap all these os calls in a try/catch -- probably not hard to imagine something going wrong, and I imagine we don't want to kill the entire agent run

mattklein (approved):

ametzger (approved):

PR #136, opened 2021-06-08
OJ-9530: Add Git diagnostic information to agent "validate" mode

daviskeene:
## Description
Currently, the agent’s “validate” mode checks that the agent can contact Jira, that it can enumerate Jira users, and that it has access to the Jira projects it’s configured for. This PR adds the following functionality:

- Ability to reach the Git instance(s) configured
- ability to see the Git orgs/project(s) configured

## Testing Strategy
Local agent "validate" mode runs
Agent output available upon request

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -167,17 +167,17 @@ def main():
                     skip_ssl_verification=config.skip_ssl_verification,
                 )
 
-                all_repos, projects = git.get_repos_from_git(client, git_config)
+                project_repo_dict = git.get_nested_repos_from_git(client, git_config)
+                all_repos = sum(project_repo_dict.values(), [])
 
-                print(
-                    f"  Agent can view the following projects for this instance: {[proj.login for proj in projects]}"
-                )
-                print(
-                    f"  Agent can view the following repos for this instance: {[repo.name for repo in all_repos]}"
-                )
+                print("  All projects and repositories available to agent:")
+                for project_name, repo_list in project_repo_dict.items():
+                    print(f"  -- {project_name}")
+                    for repo in repo_list:
+                        print(f"    -- {repo}")
 
                 for repo in git_config.git_include_repos:
-                    if repo not in [x.name for x in all_repos]:
+                    if repo not in all_repos:
                         print(f"  WARNING: {repo} is explicitly defined as an included repo, but Agent doesn't have"
                               f" proper permissions to view this repository.")
 
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -143,6 +143,32 @@ def main():
                     print(f'Error: Unable to access explicitly-included project {proj}.')
                     return
 
+        # Git diagnostics
+        import jf_agent.git as git
+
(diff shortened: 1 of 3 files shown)


jf_agent/main.py
```diff
@@ -143,6 +143,37 @@ def main():
...
+
+        for i, git_config in enumerate(git_configs, start=1):
+            print(f"Git details for instance {i}/{len(git_configs)}:")
+            print(f"  Git provider: {git_config.git_provider}")
+            print(f"  Included Projects: {git_config.git_include_projects}")
+            if len(git_config.git_exclude_projects) > 0:
+                print(f"  Excluded Projects: {git_config.git_exclude_projects}")
+            print(f"  Included Repos: {git_config.git_include_repos}")
+            if len(git_config.git_exclude_repos) > 0:
+                print(f"  Excluded Projects: {git_config.git_exclude_repos}")
```
eliwind, line 159: paste-o: should be "Excluded Repos"?

jf_agent/main.py
```diff
@@ -143,6 +143,37 @@ def main():
...
+
+            print(f'==> Testing Git connection...')
```
jeffbrainerd, line 161: totally minor but no need for the `f` string here

jf_agent/main.py
```diff
@@ -143,6 +143,37 @@ def main():
...
+                    skip_ssl_verification=config.skip_ssl_verification,
+                )
+
+                all_repos = git.get_repos_from_git(client, git_config)
+                print(
+                    f"  Agent can view the following repos for this instance: {[repo.name for repo in all_repos]}"
+                )
+            except Exception as e:
+                print("Git connection unsuccessful!")
+
```
eliwind, line 176: Could/should it do something to validate the repos?  (e.g. if there are explicitly included repos, make sure we can see them? )

jeffbrainerd (approved):

jf_agent/main.py
daviskeene: Whoops, thx

jf_agent/main.py
daviskeene: Yes, I suppose that this would be a good thing to check in the code, although we print out all repositories available so a quick glance at that list would tell you if it's visible to agent.

mattklein, line 176: It'd probably be worth also showing the Git projects the agent has access to, not just the Git repos. Maybe with the repos nested inside the projects. That'd probably take some refactoring of get_repos_from_git.

daviskeene: Grabbing the project(s) is easy enough, but it's difficult to have the repos nested inside the projects as this also requires refactoring of the `get_projects` method(s) for each adapter.

mattklein: Yeah I think that's probably worth doing though -- it'd give the best readout to the customer, especially in the not-uncommon case that there are lots of projects/repos that we have access to

mattklein (approved):
Hey awesome 🎉

ametzger (approved):

PR #137, opened 2021-06-09
[OJ-10536] Return non-zero exit code if something goes wrong

ametzger:
Return an exit code of 1 to indicate agent failure if something goes wrong during a run

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -7,6 +7,7 @@ import threading
 from collections import namedtuple
 from glob import glob
 from pathlib import Path
+import sys
 
 import requests
 import json
@@ -93,6 +94,8 @@ def main():
     creds = obtain_creds(config)
     agent_logging.configure(config.outdir)
 
+    success = True
+
     if config.run_mode == 'validate':
         print('Validating configuration...')
         from jf_jira import _get_raw_jira_connection
@@ -116,7 +119,7 @@ def main():
                 print(
                     f'Error connecting to Jira instance at {config.jira_url}, please validate your credentials. Error: {e}'
                 )
-            return
+            return False
 
         # test jira users permission
         try:
@@ -130,7 +133,7 @@ def main():
             print(
                 f'Error downloading users from Jira instance at {config.jira_url}, please verify that this user has the "browse all users" permission. Error: {e}'
             )
-            return
+            return False
 
         # test jira project access
         print('==> Testing jira project permissions...')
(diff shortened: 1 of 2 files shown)


jf_agent/main.py
```diff
@@ -178,7 +181,7 @@ def main():
                 )
                 if issues_to_scan:
                     print_missing_repos_found_by_jira(config, creds, issues_to_scan)
-                return
+                return something_went_wrong
```
eliwind, line 184: `something_went_wrong` is always `False` at this point, right?  thus far it's only ever set to `True` if its in validate mode, in which case we don't get here.  I think?

jf_agent/main.py
```diff
@@ -116,7 +119,7 @@ def main():
                 print(
                     f'Error connecting to Jira instance at {config.jira_url}, please validate your credentials. Error: {e}'
                 )
-            return
+            something_went_wrong = True
```
eliwind, line 122: Maybe a taste thing, but we *definitely* know we'll want to return `True` at this point, right?  So why not just `return True` instead of setting a variable?  The way you've written it, I have to wonder where else that variable might get set, or if it might get mutated before it's returned, or whatever else.

jf_agent/main.py
```diff
@@ -130,7 +133,7 @@ def main():
             print(
                 f'Error downloading users from Jira instance at {config.jira_url}, please verify that this user has the "browse all users" permission. Error: {e}'
             )
-            return
+            something_went_wrong = True
```
eliwind, line 136: consider `return True`

jf_agent/main.py
```diff
@@ -141,11 +144,11 @@ def main():
             for proj in config.jira_include_projects:
                 if proj not in accessible_projects:
                     print(f'Error: Unable to access explicitly-included project {proj}.')
-                    return
+                    something_went_wrong = True
```
eliwind, line 147: consider `return True`

jf_agent/main.py
```diff
@@ -141,11 +144,11 @@ def main():
...
 
         print('Success!')
 
-        return
+        return something_went_wrong
```
eliwind, line 151: If we make all the above things short-circuit, then this can `return False`

eliwind (commented):
In addition to the clarity / style comments below:

on reflection I wonder if the polarity is backwards.  Feels like a `succeeded` boolean makes more sense that a `something_went_wrong`; feels very weird to be like "everything's great, return False" and I wonder if we're going to get confused down the way.

ametzger:
@eliwind agreed, that is more clear - updated.

eliwind (approved):

PR #138, opened 2021-07-02
OJ-10908: Additional verbosity for Bitbucket Cloud

mattklein:


sarah256 (approved):
lgtm!

daviskeene (approved):

alison-j (approved):
LGTM! Was this just busted before? (That call to BitbucketCloudAdapter with no outdir specified doesn't seem like it should have worked!)

mattklein:
> LGTM! Was this just busted before? (That call to BitbucketCloudAdapter with no outdir specified doesn't seem like it should have worked!)

Yes -- though that was a bug affecting only the (relatively new) validate mode, not a general problem with pulling data

PR #143, opened 2021-07-15
OJ-11220: Add support for specifying commit branches for github

lukasbrower:
Description:
The agent assumes that only commits from the default branch of a repo should be downloaded. Customers want additional branches to be included in the download. Add a config option to specify branches for which commits should be downloaded for each repo. If no branches are given, just use the default branch as we do today.

Testing Strategy:
Ran agent locally against orthog's github. Verified that commits from a second non-develop branch are included in the json file outputs after this change. Also verified redaction still works. 

**Note**
I haven't figured out a great way to test non-github implementations. Orthog accounts for bitbucket have been removed per feedback from Alex and Eli and it is difficult to hit an existing customers instances. 

I have limited this PR to ONLY touch Github (which a requesting customer uses) but do have an untested PR which covers all implementations 
https://github.com/Jellyfish-AI/jf_agent/pull/144

Screenshots (if applicable):

### DIFF
--- a/example.yml
+++ b/example.yml
@@ -176,6 +176,14 @@ git:
   # exclude_repos:
   #    - repo_to_skip
 
+  # Uncomment this to pull commit data from each of the specified branches for each repository.
+  # By default, we will pull commit data from only the default branch of a repository.
+  # include_branches: 
+  #    - my_repository: 
+  #      - master
+  #      - develop
+
+
   # Strip out long-form text content (commit messages, PR text, etc).
   strip_text_content: False
 
--- a/example.yml
+++ b/example.yml
@@ -178,6 +178,9 @@ git:
 
   # Uncomment this to pull commit data from each of the specified branches for each repository.
   # By default, we will pull commit data from only the default branch of a repository.
+  # 
+  # Note: This is currently only supported for Github users. BitBucket and Gitlab commits
+  # will only be pulled for the default branch of a repository
   # include_branches: 
   #    - my_repository: 
   #      - master
--- a/example.yml
+++ b/example.yml
@@ -178,14 +178,15 @@ git:
 
   # Uncomment this to pull commit data from each of the specified branches for each repository.
   # By default, we will pull commit data from only the default branch of a repository.
+  # If a repo appears here with branch names which do not include the default branch, commits from
+  # the default branch will not be pulled.
   # 
   # Note: This is currently only supported for Github users. BitBucket and Gitlab commits
-  # will only be pulled for the default branch of a repository
+  # will only be pulled for the default branch of a repository.
   # include_branches: 
   #    - my_repository: 
   #      - master
   #      - develop
-
 
   # Strip out long-form text content (commit messages, PR text, etc).
   strip_text_content: False
--- a/jf_agent/git/bitbucket_cloud_adapter.py
+++ b/jf_agent/git/bitbucket_cloud_adapter.py
@@ -142,35 +142,44 @@ class BitbucketCloudAdapter(GitAdapter):
 
     @diagnostics.capture_timing()
     @agent_logging.log_entry_exit(logger)
-    def get_default_branch_commits(
-        self, normalized_repos: List[NormalizedRepository], server_git_instance_info,
+    def get_commits_for_included_branches(
+        self, normalized_repos: List[NormalizedRepository], included_branches: dict, server_git_instance_info,
     ) -> List[NormalizedCommit]:
-        print('downloading bitbucket default branch commits... ', end='', flush=True)
+        print('downloading bitbucket commits on included branches... ', end='', flush=True)
+
         for i, repo in enumerate(normalized_repos, start=1):
             with agent_logging.log_loop_iters(logger, 'repo for branch commits', i, 1):
                 pull_since = pull_since_date_for_repo(
                     server_git_instance_info, repo.project.login, repo.id, 'commits'
                 )
-                for j, commit in enumerate(
-                    tqdm(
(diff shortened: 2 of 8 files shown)


jf_agent/git/bitbucket_cloud_adapter.py
```diff
@@ -142,35 +142,44 @@ def get_repos(
...
-                            repo,
-                            self.config.git_strip_text_content,
-                            self.config.git_redact_names_and_urls,
-                        )
-                        yield commit
 
-                        # yield one commit older than we want to see
-                        if commit.commit_date < pull_since:
-                            break
+                # Find branches for which we should pull commits, specified by customer in config.
```
lukasbrower, line 156: There is some existing (and newly) duplicate code across the 4 implementations of git that we touch. I created a story to address this and have all implementations use the Adapter class that already exists. Hopefully that will let us reuse more code.
https://jelly-ai.atlassian.net/browse/OJ-11355

jf_agent/git/github.py
```diff
@@ -328,7 +345,7 @@ def _normalize_pr(client: GithubClient, pr, strip_text_content, redact_names_and
             _normalize_user(client.get_json(pr['merged_by']['url'])) if pr['merged'] else None
         ),
         commits=[
-            _normalize_commit(c, pr['base']['repo'], strip_text_content, redact_names_and_urls)
+            _normalize_commit(c, pr['base']['repo'], pr['head']['ref'], strip_text_content, redact_names_and_urls)
```
lukasbrower, line 348: Adding a branch name for the commit being created here, but I am not sure if we want the base branch or head branch's name here.

jf_agent/git/github.py
```diff
@@ -290,7 +307,7 @@ def _get_merge_commit(client: GithubClient, pr, strip_text_content, redact_names
         )
         if api_merge_commit:
             return _normalize_commit(
-                api_merge_commit, pr['base']['repo'], strip_text_content, redact_names_and_urls
+                api_merge_commit, pr['base']['repo'], pr['head']['ref'], strip_text_content, redact_names_and_urls
```
lukasbrower, line 310: Adding a branch name for the commit being created here, but I am not sure if we want the base branch or head branch's name here from the pr.

jeffbrainerd (approved):

jf_agent/git/github.py
alison-j: The past behavior would have been, all these commits have the same base, but potentially different head branches. So if we were not worrying about branch name before, it's because it was implicitly always the same, i.e. the "branch name" was the base branch. Which I think means we want base branch in the `NormalizedCommit`.

alison-j: I'm curious if that line of thought seems right to @jeffbrainerd .

alison-j (commented):
Looks good, except I think we want to be sure we're doing the right thing with head vs base in that one spot.

example.yml
```diff
@@ -176,6 +176,17 @@ git:
   # exclude_repos:
   #    - repo_to_skip
 
+  # Uncomment this to pull commit data from each of the specified branches for each repository.
```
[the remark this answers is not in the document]
lukasbrower: Good call, I'll fix this wording.

jf_agent/git/github.py
```diff
@@ -251,36 +253,51 @@ def _normalize_pr_repo(repo, redact_names_and_urls):
...
             pull_since = pull_since_date_for_repo(
                 server_git_instance_info, repo['organization']['login'], repo['id'], 'commits'
             )
-            try:
-                for j, commit in enumerate(
-                    tqdm(
-                        client.get_commits(
-                            repo['full_name'], repo['default_branch'], since=pull_since, until=None
+
+            # Find branches for which we should pull commits, specified by customer in config.
```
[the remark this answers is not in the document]
lukasbrower: Yes, thanks for catching. I am removing the second comment.

jf_agent/git/github.py
lukasbrower: The reason I went with head here is that the PRs we import should theoretically not change as they are tied to the repo rather than a specific branch. So today every PR's commits are assumed to have the same branch name and this logic would keep that behavior. Now, I can't say I have full confidence that that's what we want at the end of the day without more research/feedback into how PR commits are eventually used in Jellyfish.

I added `branch_name` to the commit objects in this PR so that we would know which branch to assign the commit to within `git_import`
https://github.com/Jellyfish-AI/jellyfish/pull/8661

So potentially the new branch_name field of commits within a PR will not even be touched later on and this is a moot point.

example.yml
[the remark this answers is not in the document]
lukasbrower: Added the following:
"If a repo appears here with branch names which do not include the default branch, commits from the default branch will not be pulled."

jf_agent/git/github.py
lukasbrower: We landed on using base here. In my previous comment I confused head and base, which makes for a very confusing back and forth...

ametzger (approved):

mattklein (commented):
Note that this should probably also include an update to the [fancy new docs](https://jellyfish-ai.github.io/jf_agent/).

lukasbrower:
> Note that this should probably also include an update to the [fancy new docs](https://jellyfish-ai.github.io/jf_agent/).

We hand most of the specific config field documentation off to the comments example file itself (which I have the description in). Happy to update if I am missing something there though
https://jellyfish-ai.github.io/jf_agent/yaml-config-setup-guide.html

PR #147, opened 2021-07-26
OJ-11482: Run Jira and Git validation only if credentials / configuration exists

daviskeene:
## Description
This PR refactors the `validate` run mode so that we can check for valid jira credentials and git configs before attempting to validate them.

## Testing
Local agent runs using the `validate` run mode and different `config.yml` configurations.


sarah256 (approved):
lgtm! I like that you put these into separate functions!

PR #151, opened 2021-08-10
OJ-11410: Allow customers to specify non-default Github branches from which to pull commits

lukasbrower:
We only pull commits from the default branch. This PR enables customers to optionally specify additional Github branches from which we should pull commits. Gitlab and Bitbucket use cases are left alone for now for the sake of PR complexity, but could be added if needed.

Verified E2E on Bitbucket Server and Github. For Github, I verified that commits from a new branch were pulled when the config is changed. All existing tests pass.


lukasbrower:
Old and obsolete

PR #155, opened 2021-09-07
OJ- 11977: Case insensitive comparisons for repo names

mattklein:
When comparing repo names against include_repos/exclude_repos from the config file, use case INsensitive comparisons.

Also a few other small bug fixes, messaging tweaks, minor code re-org.

Have tested for all four providers.

### DIFF
--- a/jf_agent/git/gitlab_adapter.py
+++ b/jf_agent/git/gitlab_adapter.py
@@ -99,6 +99,8 @@ class GitLabAdapter(GitAdapter):
             ):
                 if (
                     self.config.git_include_repos
+                    # For GitLab, git_include_repos holds IDs instead of names (probably unintentionally), so
+                    # no need to be case insensitive
                     and api_repo.id not in self.config.git_include_repos
                 ):
                     if self.config.git_verbose:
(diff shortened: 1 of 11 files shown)


jf_agent/git/gitlab_adapter.py
```diff
@@ -221,33 +231,22 @@ def get_pull_requests(
...
-
                     pull_since = pull_since_date_for_repo(
                         server_git_instance_info, nrm_repo.project.login, nrm_repo.id, 'prs'
                     )
 
                     api_prs = self.client.list_project_merge_requests(nrm_repo.id)
-                    total_api_prs = api_prs.total
-
-                    if total_api_prs == 0:
-                        agent_logging.log_and_print(
```
sarah256, line 239: Why do we no longer want to log this?

sarah256 (approved):
lgtm, just one small question!

jf_agent/git/gitlab_adapter.py
mattklein: Oh good eyes -- this was just a minor cleanup that I thought was an improvement. The new code prints these statements for each repo it's pulling from:

downloading prs for repo {name} ({id})
processing prs for {name} ({id})

Whereas the old code didn't print the first statement, just the second. (The old code used tqdm to show a progress bar progressing for each repo pulled, but didn't show the name/id of the repo currently being pulled from.) Explicitly printing out both of these lines makes it a bit easier to follow along with what the agent is doing as it's pulling PRs. So with this change, a message indicating that no PRs were found in the repo wasn't really necessary anymore; in that case there would be the "downloading" message for the repo but no "processing" message (since there would be no PRs to process).

Also the "skipping..." language used by the old message arguably wasn't really appropriate, since we usually use that term when there's an object that we try to download and parse but can't for some reason, rather than when there's legit no data to be pulled.

So just a minor improvement / judgment call -- let me know if you think any of that isn't the right call.

PR #157, opened 2021-10-01
OJ-12314 Allow pulling commits from non-default branch

lukasbrower:
Right now commits are only pulled for the default branch of specified repos. We want the option for configs to include specific branches that should be processed other than the default branch. If these are specified in the config (this is an optional section) we will only pull commits from the specified branches. 

This moves the code which enabled this to be done on Github to master, and adds support for all other implementations.

Testing:
- All unit tests pass 
- Tested E2E with Github and Gitlab to confirm that when the config option is left blank, we pull from the default branch, and when 1 or more branches are specified for a repo, we pull commits from them all.
- Code is already set up internally to process branches on commits


jf_agent/git/bitbucket_cloud_adapter.py
```diff
@@ -161,35 +161,43 @@ def get_repos(
...
-                        )
-                        yield commit
 
-                        # yield one commit older than we want to see
-                        if commit.commit_date < pull_since:
-                            break
+                # Find branches for which we should pull commits, specified by customer in config.
+                # If specific branches are not specified, just pull from default branch.
+                branches_for_repo = included_branches.get(repo.name)
+                branches = branches_for_repo if branches_for_repo else [repo.default_branch_name]
```
ametzger, line 177: I think the behavior we want is to always pull the default branch for all repos, and not force customers to specify the default branch if they specify extra branches

lukasbrower: True this may be misleading. New version always adds the default branch for a repo into the list we iterate over when fetching commits.

jf_agent/git/bitbucket_server.py
```diff
@@ -250,38 +252,43 @@ def get_default_branch_commits(
...
-                    with agent_logging.log_loop_iters(logger, 'branch commit inside repo', j, 100):
-                        if verbose:
-                            tqdm.write(
-                                f"[{datetime.now().isoformat()}] Getting {commit['id']} ({repo['name']})"
+            # Determine branches to pull commits from for this repo. If no branches are explicitly
+            # provided in a config, only pull from the repo's default branch.
+            repo_branches = [_get_default_branch_name(api_repo)]
+            additional_branches = included_branches.get(api_repo.get()['name'])
+            if additional_branches:
+                repo_branches.extend(additional_branches)
```
ametzger, line 261: any reason not to use `get_branches_for_normalized_repo` here?

jf_agent/git/utils.py
```diff
@@ -0,0 +1,10 @@
+from jf_agent.git import NormalizedRepository
+
+# Return branches for which we should pull commits, specified by customer in git config.
+# The repo's default branch will always be included in the returned list.
+def get_branches_for_normalized_repo(repo: NormalizedRepository, included_branches: dict):
+    repo_branches = [repo.default_branch_name]
+    additional_branches_for_repo = included_branches.get(repo.name)
+    if additional_branches_for_repo:
+        repo_branches.extend(additional_branches_for_repo)
+    return repo_branches
```
ametzger, line 10: nit: might want to do `return set(repo_branches) here to eliminate dupes

jf_agent/git/bitbucket_server.py
lukasbrower: This is because of an annoying heterogeneity between our implementations of BitBucket Server and Github compared to Gitlab and Bitbucket cloud. 

The latter use adapters which make it so that we are already working with our internal, normalized data structures. The former directly interact with the API results so the fields we care about (repo name in this case) are retrieved differently across APIs. Ideally we would move everything to the adapter pattern for consistency.

jf_agent/git/utils.py
lukasbrower: Good idea

jf_agent/git/bitbucket_server.py
ametzger: heterogeneity is a great word, that makes sense. I might just add a comment here for future refactorers (refactorinos?) so they don't attempt to simplify and break stuff.

💯 % agreed about cleaning up our adapter pattern usage here.

jeffbrainerd (approved):

ametzger (approved):

PR #161, opened 2021-10-18
[OJ-12496] Add bearer token authorization for Jira Server

ametzger:
In cases where a customer has disabled basic authentication on Jira server (see [here](https://confluence.atlassian.com/enterprise/disabling-basic-authentication-1044776464.html)) we want to allow using `Authorization: Bearer <token>` headers directly instead of forcing all users to use basic auth. This is a bit hacky but will allow customers to use bearer tokens in the short term.

Tested this locally against a Jira server running in Docker on which I disabled basic authentication and against the Jellyfish cloud Jira instance, both of which worked.

Patch has also been submitted upstream, so we should use that if/when it is merged.

### DIFF
--- a/docs/_setups/jira-setup.md
+++ b/docs/_setups/jira-setup.md
@@ -2,6 +2,10 @@
 title: Jira Setup
 ---
 
+Before you begin, determine which credential setup you will need to use. Most of the time, you will want to set up credentials using basic auth. If you are running Jira server (i.e. not Jira cloud) and have [disabled basic authentication](https://confluence.atlassian.com/enterprise/disabling-basic-authentication-1044776464.html), you will want to use bearer token auth. If you aren't sure, use basic auth.
+
+## Basic Auth
+
 1. Add the following section to your environment variable file. This is the same file mentioned in step 3 above. Adding the following variables allows the agent to access your Jira data:
     <p class="code-block"><code>
         JIRA_USERNAME=...<br/>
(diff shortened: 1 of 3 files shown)


docs/.ruby-version
```diff
@@ -0,0 +1 @@
+3.0.1
```
eliwind, line 1: What's this (and the webrick dependency below) coming from?  I'm surprised editing doc introduces a new dependency, and it looks like webrick is an HTTP server?

jf_agent/main.py
```diff
@@ -248,21 +256,24 @@ def obtain_creds(config):
...
     # obtain git slug to credentials
     git_instance_to_creds = {
         git_config.git_instance_slug: _get_git_instance_to_creds(git_config)
         for git_config in config.git_configs
     }
 
-    if config.jira_url and not (jira_username and jira_password):
+    # This is a bit hairy, they must either provide a jira username
+    # and a jira password OR a single jira bearer token.
+    if config.jira_url and ((not (jira_username and jira_password)) and (not jira_bearer_token)):
```
eliwind, line 269: `not ((jira_username and jira_password) ^ jira_bearer_token)` perhaps?

docs/_setups/jira-setup.md
```diff
@@ -13,3 +17,16 @@ title: Jira Setup
...
+
+## Bearer Token Auth
+
+1. Create a user in your Jira server that has read access to all projects you would like the agent to ingest. If you have already created such a user, open a browser while logged in as said user.
+
+2. Retrieve a Personal Access Token for the user from step 2 following [this guide](https://confluence.atlassian.com/enterprise/using-personal-access-tokens-1026032365.html).
+
+3. Add the following section to your environment variable file. This is the same file mentioned in step 3 above. Adding the following variables allows the agent to access your Jira data:
+    <p class="code-block"><code>
+        JIRA_BEARER_TOKEN=...<br/>
```
sarah256, line 29: nit (feel free to ignore) I don't think you need a `<br/>` here

jf_agent/main.py
sarah256: This makes sense but also might be a bit less readable (unless this way is more pythonic?)

sarah256 (approved):
lgtm, just some small comments!

docs/.ruby-version
eliwind: is there such a thing as dev dependencies in there then?  (make it clear it's local dev stuff not something we need at runtime).

PR #167, opened 2021-12-16
fix(gitlab): handle exception on group api call

dinoDayo:
`jf_agent` logging improvements incorporated and tested. Thank you!


Pipfile.lock
```diff
@@ -1,412 +0,0 @@
-{
```
mattklein, line 1: You're erroneously deleting this file -- we don't want to do that!

You should be able to get it back with:
```
git reset HEAD~1
git restore Pipfile.lock
git add .
git commit --amend
git push --force-with-lease
```

mattklein (changes requested):

mattklein (approved):

PR #168, opened 2022-01-25
OJ-12951: Allow regex branch names in config

lukasbrower:
This adds the ability for users to use wildcard patterns when specifying the branches from which we should pull commits. 

### Testing
Ran against Github and Gitlab implementations and verified that the expected branches were pulled when using wildcards

### DIFF
--- a/jf_agent/git/utils.py
+++ b/jf_agent/git/utils.py
@@ -5,12 +5,12 @@ from jf_agent.git import NormalizedRepository
 # Return branches for which we should pull commits, specified by customer in git config.
 # The repo's default branch will always be included in the returned list.
 def get_branches_for_normalized_repo(repo: NormalizedRepository, included_branches: dict):
-    included_branches = [repo.default_branch_name]
+    branches_to_process = [repo.default_branch_name]
     additional_branches_for_repo = included_branches.get(repo.name)
     if additional_branches_for_repo:
         repo_branch_names = [b.name for b in repo.branches]
-        included_branches.extend(get_matching_branches(additional_branches_for_repo, repo_branch_names))
-    return set(included_branches)
+        branches_to_process.extend(get_matching_branches(additional_branches_for_repo, repo_branch_names))
+    return set(branches_to_process)
 
 # Given a list of patterns, either literal branch names or regexes meant to match a set of branches in a repo, 
 # return the list of branches from repo_branches match any of the patterns.
--- a/jf_agent/git/utils.py
+++ b/jf_agent/git/utils.py
@@ -1,10 +1,22 @@
+import re
+from typing import List
 from jf_agent.git import NormalizedRepository
 
 # Return branches for which we should pull commits, specified by customer in git config.
 # The repo's default branch will always be included in the returned list.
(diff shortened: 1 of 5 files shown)


jf_agent/git/utils.py
```diff
@@ -1,10 +1,22 @@
...
+        repo_branch_names = [b.name for b in repo.branches]
+        branches_to_process.extend(get_matching_branches(additional_branches_for_repo, repo_branch_names))
+    return set(branches_to_process)
+
+# Given a list of patterns, either literal branch names or regexes meant to match a set of branches in a repo, 
+# return the list of branches from repo_branches that match any of the branch name patterns.
+def get_matching_branches(included_branch_patterns: List[str], repo_branch_names: List[str]) -> List[str]:
+    matching_branches = []
+    for repo_branch_name in repo_branch_names:
+        if any(re.compile(pattern).fullmatch(repo_branch_name) for pattern in included_branch_patterns):
```
ametzger, line 20: nit: no need to do `re.compile` here, that is mostly useful for regexes you're going to be reusing a lot:
``` python
if any(re.fullmatch(pattern, repo_branch_name) for pattern in included_branch_patterns):
```

ametzger (approved):
one nit, otherwise LGTM

jf_agent/git/utils.py
lukasbrower: Good call, removed

eliwind (commented):
This looks reasonable, but I wonder if regex is overkill here.  Given the relative unsophistication of our agent users, and that the one use case we have is a simple wildcard match, I winder if we ought to have less footgun and use some simpler wildcard syntax rather than full regex.

On the other hand, I don't immediately know of a starnard wildcard syntax or library for matching...

lukasbrower:
> This looks reasonable, but I wonder if regex is overkill here. Given the relative unsophistication of our agent users, and that the one use case we have is a simple wildcard match, I winder if we ought to have less footgun and use some simpler wildcard syntax rather than full regex.
> 
> On the other hand, I don't immediately know of a starnard wildcard syntax or library for matching...

I had similar thoughts, the main reason I opted for this was to avoid introducing our own custom wildcard matching logic and offload that to traditional regex parsing.

We could just assume all '*' mean wildcard as the ticket suggests but I worry custom syntax might confuse customers more than universally known regex rules

lukasbrower:
Talked with Team Jo and we agreed it is probably best to limit this to just wildcards which is possible with the fnmatch library (meant to unix file patterns, but the functionality works generally on strings and only features the use of * and ? special characters)

Updated this PR to reflect that

dregalbuto (approved):

alison-j (approved):
Looks good. For paranoia I will ask, did the test config include any NON-wildcards? (i.e. did we confirm the old list-of-fixed-strings works?)

lukasbrower:
> Looks good. For paranoia I will ask, did the test config include any NON-wildcards? (i.e. did we confirm the old list-of-fixed-strings works?)

Yes! I tested the base case as well

PR #171, opened 2022-02-09
Add init file to test directory to enable unittest discovery

lukasbrower:
unittest discovery only looks in packages (directory containing a file named `__init__.py`). Add one to the test directory

Before:
`pipenv run python -m unittest tests/test_*.py`

After:
`pipenv run python -m unittest`


chtinahoward (approved):
👍

PR #172, opened 2022-02-10
OJ-14076-keyloop-agent-error-not-enough-values-to-unpack-expected-2-got-0-poc

phalaphone:
Ticket: https://jelly-ai.atlassian.net/browse/OJ-14076

Upon investigation it was found that they "not enough value" error was caused by bad bitbucket configuration. This update causes validation and a run to log the error and continue on (so multiple errors can be surfaced in a run and addressed at once)

### DIFF
--- a/jf_agent/validation.py
+++ b/jf_agent/validation.py
@@ -87,29 +87,35 @@ def validate_git(config, creds):
             project_repo_dict = get_nested_repos_from_git(client, git_config)
             all_repos = sum(project_repo_dict.values(), [])
 
-            print("  All projects and repositories available to agent:")
-            for project_name, repo_list in project_repo_dict.items():
-                print(f"  -- {project_name}")
-                for repo in repo_list:
-                    print(f"    -- {repo}")
+            if not all_repos:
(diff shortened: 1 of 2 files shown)


jf_agent/validation.py
```diff
@@ -87,29 +87,35 @@ def validate_git(config, creds):
...
-                if comp_func(repo):
-                    print(
-                        f"  WARNING: {repo} is explicitly defined as an included repo, but agent doesn't seem"
-                        f" to see this repository -- possibly missing permissions."
-                    )
+            if not all_repos:
+                print(
+                    " =============================================================================/n \033[91mERROR: No projects and repositories available to agent: Please Check Configuration\033[0m /n ============================================================================="
+                )
+
```
mattklein, line 94: It's (arguably) more readable to have the `if` block do a `continue` after the `print`. Then you wouldn't need the `else` and you could de-indent the code that's inside of it back to the level it was previously at.

mattklein (approved):
A couple of minor comments, lgtm!

PR #178, opened 2022-05-12
doc: document the use of REQUESTS_CA_BUNDLE to work with custom certificates

olivierdagenais:
Rather than use the instance-wide `no_verify_ssl` switch, these changes document how to configure the agent to trust more certificates than those that came with the application.

# Manual testing

Given:
1. a Bash script called `agent.sh` similar to the snippet in the `README.md`
2. a configuration file which contained `no_verify_ssl: False` and 2 Git configurations
3. a `cacert.pem` file which was augmented with some CA certificates our administrators are using to issue certificates for various services

When:
I ran `./agent.sh -m validate`

Then:
1. The first Git entry in our configuration file is our self-hosted Bitbucket server, which has a certificate issued by a custom CA certificate.  The connection was successful. (without these changes, this would fail to connect with the `CERTIFICATE_VERIFY_FAILED` error)
2. The 2nd Git entry in our configuration file is an organization in GitHub Cloud, which also succeeds validation, showing that the certificate bundle was indeed _augmented_.

Mission accomplished!

### DIFF
--- a/example.yml
+++ b/example.yml
@@ -3,8 +3,10 @@
 #############################
 
 global:
-  # Set this to True to skip verification of server SSL certificates.
-  # This option is deprecated; see the README.md for how to trust custom certificates.
+  # Set this to True to skip verification of server SSL certificates.  This might
+  # be necessary if your Jira / Bitbucket server has an invalid or a custom
+  # certificate.  Or as an alternative, see the README.md for how to trust custom
+  # certificates.
   no_verify_ssl: False
   # Set this to True if you would like to send your config.yml file to Jellyfish.
   # We use this as a diagnostic tool for customer assistance.
--- a/example.yml
+++ b/example.yml
@@ -3,8 +3,8 @@
 #############################
 
 global:
-  # Set this to True to skip verification of server SSL certificates.  This might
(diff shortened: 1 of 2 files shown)


eliwind (commented):
Thanks!  We'll confirm these instructions, but agree this seems directionally like a good thing to document.

example.yml
```diff
@@ -3,8 +3,8 @@
 #############################
 
 global:
-  # Set this to True to skip verification of server SSL certificates.  This might
-  # be useful if your Jira / Bitbucket server doesn't have a valid SSL certificate.
+  # Set this to True to skip verification of server SSL certificates.
+  # This option is deprecated; see the README.md for how to trust custom certificates.
```
mattklein, line 7: I might ask that we not go so far as to say it's "deprecated"... we don't really want to sound prescriptive on this to current or future customers :)

mattklein (commented):
Hey this is great, glad you were able to do that and thanks for putting these instructions together! I'm sure they will be useful to people down the road.

I might ask for one wording tweak re: "deprecation"...

mattklein (approved):

example.yml
olivierdagenais: Thanks for adjusting!

PR #184, opened 2022-07-07
Cache Organization details

acicchetti-jf:
Organizations are unlikely to change during the run of the agent, so it should be fairly safe to keep them cached. Ideally this reduces load on Github for large numbers of repositories in the same organization


mattklein (approved):

acicchetti-jf:
I think so - there doesn't seem to be any appetite for the change (anymore), and it's not thoroughly tested enough that we should put it in willy-nilly

PR #193, opened 2022-09-30
OJ-18051: Chunk up long project_ids for GET request

zdwarren:
A customer is having a problem of the GET url being too long.
This ticket will chuck on the project_ids if they are too long.

Tested locally with ortho.

Jira Ticket:
https://jelly-ai.atlassian.net/browse/OJ-18051

### DIFF
--- a/jf_agent/jf_jira/jira_download.py
+++ b/jf_agent/jf_jira/jira_download.py
@@ -290,89 +290,106 @@ IssueMetadata = namedtuple('IssueMetadata', ('key', 'updated'))
 @diagnostics.capture_timing()
 @agent_logging.log_entry_exit(logger)
 def download_all_issue_metadata(
-    jira_connection, project_ids, earliest_issue_dt, num_parallel_threads, issue_filter
+    jira_connection, all_project_ids, earliest_issue_dt, num_parallel_threads, issue_filter
 ) -> Dict[int, IssueMetadata]:
     print('downloading issue metadata... ', end='', flush=True)
-    issue_jql = (
-        f'project in ({",".join(project_ids)}) and updatedDate > '
-        f'{"0" if not earliest_issue_dt else earliest_issue_dt.strftime("%Y-%m-%d")}'
-    )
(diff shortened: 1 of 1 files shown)


jf_agent/jf_jira/jira_download.py
```diff
@@ -290,89 +290,106 @@ def download_boards_and_sprints(jira_connection, project_ids, download_sprints):
...
     print('downloading issue metadata... ', end='', flush=True)
-    issue_jql = (
-        f'project in ({",".join(project_ids)}) and updatedDate > '
-        f'{"0" if not earliest_issue_dt else earliest_issue_dt.strftime("%Y-%m-%d")}'
-    )
-    if issue_filter:
-        issue_jql += f' and {issue_filter}'
-    total_num_issues = jira_connection.search_issues(issue_jql, fields=['id']).total
-    issues_per_thread = math.ceil(total_num_issues / num_parallel_threads)
+
```
zdwarren, line 296: This is the section with the changes - the diff is confusing because the for loop changed the spacing for everything below it.

eliwind (commented):
This seem like a valid way to chunk up GET requests.  I'm curious though about GET for POST: there is another place (where we download the full issues rather than just the update timestamps) where we make the same requests with POST instead to avoid this issue.

Is there a reason this part ought to use GET / be different?

I'm not trying to block merging here, just curious to understand the thinking better.

eliwind:
Also worth looking at the direct-connect version of this code too -- whatever we do in the agent we should probably do there as well.

This looks fine to me, I leave the relative merits of going to POST for consistency with the other place we make this call vs leaving it as a GET for less risk perturbing existing customers to your call.

Other thought: is there reason to think _think_ URL limit is the right one?  Will other load balances / HTTP servers / whatever else have a smaller limit (and thus maybe we should chunk even smaller?)

eliwind (approved):

mattklein (approved):

PR #196, opened 2022-10-26
OJ-18980: project_ids bug: Change project_ids to list

zdwarren:
Needs to be a list so it can be split - comes in as a set.


laurado (approved):

chtinahoward (approved):
👍

PR #201, opened 2023-01-18
Parallelize git download by git configuration

jduffymclaughlin:
This adds basic parallelization to git downloading, by git configuration. Each git config is subject to a different rate-limit so this should be a substantial speed performance for multi-git use-cases.

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -370,15 +370,18 @@ def download_data(config, creds, endpoint_jira_info, endpoint_git_instances_info
             print_all_jira_fields(config, jira_connection)
         download_data_status.append(load_and_dump_jira(config, endpoint_jira_info, jira_connection))
 
-    
+    if len(config.git_configs) == 0: 
+        return download_data_status
+
     # git downloading is parallelized by the number of configurations. 
     futures = []
-    with ThreadPoolExecutor(max_workers=8) as executor:
+    with ThreadPoolExecutor(max_workers=config.git_max_concurrent) as executor:
         for git_config in config.git_configs:
             agent_logging.log_and_print(
                 logger,
                 logging.INFO,
-                f'Obtained {git_config.git_provider}:{git_config.git_instance_slug} configuration, attempting download {"in parallel" if len(config.git_configs) > 1 else ""}...',
+                f'Obtained {git_config.git_provider}:{git_config.git_instance_slug} configuration, attempting download '
+                + f'in parallel with {config.git_max_concurrent} workers' if len(config.git_configs) > 1 else "..."
             )
             futures.append(
                 executor.submit(
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -1,4 +1,5 @@
 import argparse
+from concurrent.futures import ThreadPoolExecutor
 import gzip
 import logging
 import os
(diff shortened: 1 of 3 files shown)


jf_agent/main.py
```diff
@@ -366,37 +370,53 @@ def download_data(config, creds, endpoint_jira_info, endpoint_git_instances_info
...
-            instance_info = endpoint_git_instances_info.get(instance_slug)
-            instance_creds = creds.git_instance_to_creds.get(instance_slug)
-        else:
-            # support legacy single-git support, which assumes only one available git instance
-            instance_info = list(endpoint_git_instances_info.values())[0]
-            instance_creds = list(creds.git_instance_to_creds.values())[0]
+    
+    # git downloading is parallelized by the number of configurations. 
+    futures = []
+    with ThreadPoolExecutor(max_workers=8) as executor:
```
mattklein, line 376: I think we should probably make max_workers be parametrizable, with the ability to override (set it to something smaller than the default) it in the config file. With the default being the min of (number of instances defined in the config file) or (something non-crazy like 8?).

jf_agent/main.py
```diff
@@ -366,37 +370,53 @@ def download_data(config, creds, endpoint_jira_info, endpoint_git_instances_info
...
+                    creds,
+                    endpoint_git_instances_info,
+                    len(config.git_configs) > 1,
+                )
+            )
 
-        git_connection = get_git_client(
-            git_config, instance_creds, skip_ssl_verification=config.skip_ssl_verification
-        )
+    return download_data_status + [f.result() for f in futures]
```
mattklein, line 394: If you haven't already, I'd double-check that the exception handling is what you'd want it to be -- i.e., if an exception propagates out of one of the threads, it's raised in the main thread (rather than being swallowed).

jf_agent/main.py
```diff
@@ -366,37 +370,53 @@ def download_data(config, creds, endpoint_jira_info, endpoint_git_instances_info
...
+) -> dict:
+
+    if is_multi_git_config:
+        instance_slug = git_config.git_instance_slug
+        instance_info = endpoint_git_instances_info.get(instance_slug)
+        instance_creds = creds.git_instance_to_creds.get(instance_slug)
+    else:
+        # support legacy single-git support, which assumes only one available git instance
+        instance_info = list(endpoint_git_instances_info.values())[0]
+        instance_creds = list(creds.git_instance_to_creds.values())[0]
```
mattklein, line 408: Just want to make sure this legacy path has been tested too (refactor looks right to my eye though)

jf_agent/main.py
jduffymclaughlin: The `concurrent.Future.result()` method will raise the exception if one occurred on the thread. 
https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result
I tried this out by throwingn in `_download_git_data` and it works as expected.

jf_agent/main.py
```diff
@@ -370,15 +370,18 @@ def download_data(config, creds, endpoint_jira_info, endpoint_git_instances_info
             print_all_jira_fields(config, jira_connection)
         download_data_status.append(load_and_dump_jira(config, endpoint_jira_info, jira_connection))
 
-    
+    if len(config.git_configs) == 0: 
+        return download_data_status
```
jduffymclaughlin, line 374: `ThreadPoolExecutor(max_workers=0)` throws a ValueError so I decided to short-circuit here

jf_agent/main.py
jduffymclaughlin: https://github.com/Jellyfish-AI/jf_agent/pull/201/commits/ef02b7c0d3264832cc3fe67037efc4754bc60344

jf_agent/main.py
jduffymclaughlin: 👍  I have tested configs of the form: 
```
git:
  provider: github
  url: https://api.github.com
  ....
```
and those of the form 
```
git:
  - provider: github
    url: https://api.github.com
    ...
  - provider: gitlab
     url: gitlab.com
    ...
``` 
which hit either of these cases.

mattklein (approved):

PR #205, opened 2023-02-08
OJ-21219-better-logging-in-agent-submission

gavinpitt-jf:
Improve logging to the agent uploading logic to ensure that all logs, and possible exceptions, are captured.

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -526,16 +526,16 @@ def send_data(config, creds):
         config_file_dict = get_signed_url(['config.yml'])['config.yml']
         upload_file('config.yml', config_file_dict['s3_path'], config_file_dict['url'], local=True)
 
+    # Upload log files as last step before uploading the .done file
+    log_file_dict = get_signed_url([agent_logging.LOG_FILE_NAME])[agent_logging.LOG_FILE_NAME]
+    upload_file(agent_logging.LOG_FILE_NAME, log_file_dict['s3_path'], log_file_dict['url'])
+
     if success:
         # creating .done file, only on success
         done_file_path = f'{os.path.join(config.outdir, ".done")}'
         Path(done_file_path).touch()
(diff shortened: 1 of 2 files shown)


jf_agent/main.py
```diff
@@ -499,6 +505,11 @@ def upload_file(filename, path_to_obj, signed_url, local=False):
         t.join()
 
     if any(thread_exceptions):
+        # Run through exceptions and inject them into the agent log
+        for exception in thread_exceptions:
+            agent_logging.log_and_print_error_or_warning(
+                logger, logging.ERROR, error_code=1000, msg_args=[exception], exc_info=True
```
mattklein, line 511: I think you mean to use the code `0000`?

gavinpitt-jf: Oh, right. Good catch

jf_agent/main.py
```diff
@@ -498,24 +509,35 @@ def upload_file(filename, path_to_obj, signed_url, local=False):
...
+        done_file_path = f'{os.path.join(config.outdir, ".done")}'
+        Path(done_file_path).touch()
+        done_file_dict = get_signed_url(['.done'])['.done']
+        upload_file('.done', done_file_dict['s3_path'], done_file_dict['url'])
 
-    return True
+    # Upload log files as last step
+    log_file_dict = get_signed_url([agent_logging.LOG_FILE_NAME])[agent_logging.LOG_FILE_NAME]
+    upload_file(agent_logging.LOG_FILE_NAME, log_file_dict['s3_path'], log_file_dict['url'])
+
```
mattklein, line 539: Actually I would make the `.done` file be the very last file uploaded, because the existence of that file is what triggers us to begin server-side processing. So if the `.done` file is uploaded before the log file I think it's possible we'll start to process the submission and miss the log file. So I think the last ones should be the log file and then the `.done` file.

mattklein (approved):

PR #206, opened 2023-02-09
OJ-21409-add-pre-commit-formatting-rules

gavinpitt-jf:
Add `black` formatting logic to the `jf_agent` repository, so that all code stays consistently formatted.

I opted to lock us into `black` version `19.10b0` to keep us in sync with other internal repositories. Locking us into that version required us to install a dependency on the `click` library, which I also locked into place to ensure we stay in sync with `black`. Additionally I need to install the `pre-commit` library. This is all contained within the python virtual env.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+repos:
+  - repo: https://github.com/ambv/black
+    rev: 19.10b0
+    hooks:
+      - id: black
+        language_version: python3.9
+        additional_dependencies: ["click==8.0.4"]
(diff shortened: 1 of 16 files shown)


CaptainKanuk (changes requested):
Good change - how are we going to test it to make sure it still works when deployed to customer servers?

gavinpitt-jf:
> Good change - how are we going to test it to make sure it still works when deployed to customer servers?

We could test that on our local machines, I bet. I need to brush up on how code is pulled and run from the agent, and if messing with the pipfile is going to affect how the code is run

mattklein (approved):
Looks good to me. With all the dependency changes I'd just keep an eye on agent submissions after this gets deployed and make sure the new version seems to not be problematic for any customer.

PR #210, opened 2023-03-03
#OJ-22145 large files uploads intermittently fail under certain network conditions, batch data into smaller files

jobelenus:


jf_agent/__init__.py
```diff
@@ -34,6 +37,15 @@ def default(self, o):
...
 
 
+def _batched(iterable, n):
+    "Batch data into tuples of length n. The last batch may be shorter."
+    # batched('ABCDEFG', 3) --> ABC DEF G
+    if n < 1:
+        raise ValueError('n must be at least one')
+    it = iter(iterable)
+    while (batch := tuple(islice(it, n))):
+        yield batch
```
ametzger, line 47: nit: I might move this into `jf_agent.utils`, which has another similar function https://github.com/Jellyfish-AI/jf_agent/blob/fac7c6212898e7c3e786b8623a77f28444915716/jf_agent/util.py#L4

ametzger (approved):

jf_agent/__init__.py
jobelenus: Right... I worry that the `split` won't work with a generator. But definitely will move this

PR #213, opened 2023-03-15
OJ-22511, OJ-22619: skip issue on 500

thingscouldbeworse:
Also OJ-22619

### DIFF
--- a/jf_agent/jf_jira/jira_download.py
+++ b/jf_agent/jf_jira/jira_download.py
@@ -310,6 +310,47 @@ def download_boards_and_sprints(jira_connection, project_ids, download_sprints):
 IssueMetadata = namedtuple('IssueMetadata', ('key', 'updated'))
 
 
+def get_issues(jira_connection, issue_jql, start_at, batch_size):
+    try:
+        api_response = jira_connection.search_issues(
+            f'{issue_jql} order by id asc',
+            fields=['updated'],
+            startAt=start_at,
+            maxResults=batch_size,
+        )
+    except (JIRAError, KeyError) as e:
+        if hasattr(e, 'status_code') and e.status_code < 500:
+            # something wrong with our request; re-raise
+            raise
+
+        # We have seen sporadic server-side flakiness here. Sometimes Jira Server (but not
+        # Jira Cloud as far as we've seen) will return a 200 response with an empty JSON
+        # object instead of a JSON object with an "issues" key, which results in the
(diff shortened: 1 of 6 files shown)


jf_agent/jf_jira/jira_download.py
```diff
@@ -384,10 +384,14 @@ def _download_some(thread_num, start_at, end_at):
...
-                                logger, logging.ERROR, error_code=3022,
+                            # copied logic from jellyfish direct connect
+                            # don't bail, just skip
+                            # khardy 2023-03-16
+                            agent_logging.log_and_print_error_or_warning(logger, logging.WARNING, msg_args=
+                                f"Caught {type(e)} from search_issues(), batch size is already 0, giving up on "
+                                "fetching this issue's metadata", error_code=3092
                             )
-                            raise
+                            start_at += 1
```
jruel4, line 394: Q: Should we have a "continue" here?

mattklein, line 394: Also I think at this point we'll have reduced batch size to 0, so on the next iteration of the loop the batch size will still be 0; so I'm not sure that subsequent iterations are going to actually fetch the rest of the issues. Resetting the batch size back up to its starting point may be necessary. 🤔

jf_agent/jf_jira/jira_download.py
```diff
@@ -356,45 +397,17 @@ def _download_some(thread_num, start_at, end_at):
...
+                    api_resp = get_issues(jira_connection, issue_jql, start_at, batch_size)
 
                     issue_metadata = {
                         int(iss.id): IssueMetadata(iss.key, parser.parse(iss.fields.updated))
                         for iss in api_resp
                     }
                     all_issue_metadata.update(issue_metadata)
-                    start_at += len(issue_metadata)
+                    if len(issue_metadata) == 0:
+                        start_at += 1
```
mattklein, line 408: This could probably use a comment, since it's not clear from the line itself what the rationale is for this. Something like "get_issues reduced batch size all the way down to 1, still didn't successfully fetch the single issue, so we're skipping it"

mattklein (approved):
Looks good to me, just one suggestion that you can take or leave.

jf_agent/jf_jira/jira_download.py
thingscouldbeworse: good call, added here: https://github.com/Jellyfish-AI/jf_agent/pull/213/commits/9211a053e9540760a017bdc111ce382f6ad91b67

PR #217, opened 2023-04-14
[OJ-23618] Add precommit and CI job to detect secrets in PRs or commits

JosiahOne:
This PR adds a new pre-commit hook and GitHub actions job to detect accidental inclusion of sensitive data (private keys, slack webhooks, etc.)

A "baseline" file is included which will prevent detect-secrets from reporting on findings in those data files in the future. I audited all those findings and they were not actual issues (they were git commit hashes).

**Testing:**

1. Commit and push change locally (when your change does not have any secrets included in it).
2. Note pre-commit does not flag any issues
3. Note CI passes cleanly
4. Now create fake change which includes a "suspicious" value. i.e. include `API_KEY = 'Test'` somewhere.
5. Commit and push
6. Note pre-commit fails
7. Push with `--no-verify` to skip pre-commit hooks
8. Note that CI Job fails.

**Screenshots:**

Example of pre-commit failure:
Screenshot 2023-04-14 at 10 36 44 AM

Example of CI failure:
Screenshot 2023-04-14 at 10 38 18 AM

### DIFF
--- a/.github/workflows/secrets-detect.yml
+++ b/.github/workflows/secrets-detect.yml
@@ -1,0 +1,17 @@
+
+name: Secrets detection using detect-secrets
+on:
+  pull_request:
+    branches: [ "master" ]
+jobs:
+  secrets-detect:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
+          fetch-depth: 0
+      - name: Install detect-secrets
+        run: pip install detect-secrets
+      - name: Run detect secrets.
+        run: git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | grep -v 'Pipfile.lock' | xargs detect-secrets-hook
+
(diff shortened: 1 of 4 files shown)


.github/workflows/secrets-detect.yml
```diff
@@ -0,0 +1,17 @@
+
+name: Secrets detection using detect-secrets
+on:
+  pull_request:
+    branches: [ "master" ]
```
IamParadoxdotexe, line 5: I don't think we actually use the `master` branch of `jellyfish`? Should this be `develop`?

IamParadoxdotexe (approved):
1 q

.github/workflows/secrets-detect.yml
JosiahOne: Agreed! But this is `jf_agent` where we don't use develop. :)

PR #219, opened 2023-05-02
OJ-24273-agent-manifest-upload-cleanup

gavinpitt-jf:
There were a few bugs that arose from the manifest uploading logic that shipped yesterday. The bugs are isolated to ONLY manifest uploading, so normal agent uploads were not affected.

The following bugs have been fixed with this PR:
- Previously git Instance slugs were getting set to None if the `instance_slug` was not set in the`config.yaml` file (this is common)
  - Instance slugs are now pulled from the `endpoint_info` dictionary, which comes from a Jellyfish API endpoint (this is how the git processing works for agent in general)
- Getting the Jira Users Count could potentially error out for some obscure Jira Instances
  - Getting the Jira Users count now leverages the "get_users" function that already exists in the `jira_downloader` module. This should cover every use case we've encountered so far
- I broke Jira and Git manifest generation into separated try catch blocks, which should allow us to upload git manifests in the case where something goes wrong with uploading jira manifests

### DIFF
--- a/jf_agent/data_manifests/git/generator.py
+++ b/jf_agent/data_manifests/git/generator.py
@@ -1,4 +1,5 @@
 import logging
+from jf_agent import agent_logging
 from jf_agent.config_file_reader import GitConfig
 
 from jf_agent.data_manifests.git.adapters.manifest_adapter import ManifestAdapter
@@ -31,8 +32,10 @@ def create_manifests(
             user_manifests: list[GitUserManifest] = []
             team_manifests: list[GitTeamManifest] = []
             for org in git_config.git_include_projects:
-                logger.info(
-                    f'Processing git instance {git_config.git_instance_slug} for company {company_slug} under github org {org}'
+                agent_logging.log_and_print(
+                    logger,
+                    logging.INFO,
+                    f'Processing git instance {git_config.git_instance_slug} for company {company_slug} under github org {org}',
                 )
 
                 instance_creds = creds.git_instance_to_creds.get(git_config.git_instance_slug)
@@ -45,8 +48,10 @@ def create_manifests(
                 )
 
                 repos_count = manifest_adapter.get_repos_count()
-                logger.info(
-                    f'Processing {repos_count} repos {"including Branches and Pull Request" if verbose else ""}'
+                agent_logging.log_and_print(
+                    logger,
+                    logging.INFO,
+                    f'Processing {repos_count} repos {"including Branches and Pull Request" if verbose else ""}',
                 )
                 for repo_manifest in manifest_adapter.get_all_repo_data():
                     if verbose:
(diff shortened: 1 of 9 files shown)


jf_agent/main.py
```diff
@@ -392,23 +394,48 @@ def generate_manifests(config, creds):
...
+            )
+            # Create and add Git Manifests
+            manifests += create_git_manifests(
+                company_slug=company_slug,
+                creds=creds,
+                config=config,
+                endpoint_git_instances_info=jellyfish_endpoint_info.git_instance_info,
+            )
+        except Exception as e:
+            agent_logging.log_and_print(
```
jruel4, line 434: NB: Consider adding `traceback.format_exc()` here, so that we have exact details on what's erroring

jf_agent/main.py
```diff
@@ -392,23 +394,48 @@ def generate_manifests(config, creds):
...
+                manifests.append(jira_manifest)
+            else:
+                agent_logging.log_and_print(
+                    logger, logging.ERROR, 'create_jira_manifest returned a None Type.'
+                )
+        except Exception as e:
             agent_logging.log_and_print(
-                logger, logging.ERROR, 'create_jira_manifest returned a None Type.'
+                logger,
+                logging.ERROR,
```
jruel4, line 413: NB: Same as https://github.com/Jellyfish-AI/jf_agent/pull/219/files#r1182716687

jf_agent/data_manifests/jira/generator.py
```diff
@@ -17,25 +17,37 @@ class UnsupportedJiraProvider(Exception):
...
 
-    agent_logging.log_and_print(
-        logger, logging.INFO, f'Generating Jira Manifest data for {company_slug}...'
-    )
+    def _agent_log(msg: str):
+        agent_logging.log_and_print(logger, logging.INFO, msg)
+
+    _agent_log(f'Generating Jira Manifest data for {company_slug}...')
 
+    _agent_log('Processing Users...')
```
jruel4, line 25: C: Nice addition.

jf_agent/data_manifests/git/generator.py
```diff
@@ -104,32 +123,31 @@ def create_manifests(
...
+                        repo_manifests=repo_manifests,
+                        user_manifests=user_manifests,
+                        team_manifests=team_manifests,
+                    )
                 )
-            )
         except UnsupportedGitProvider as e:
-            logger.error(
-                f'An exception happened when creating {type} manifest for {git_config.git_instance_slug}. Err: {e}'
+            agent_logging.log_and_print(
```
jruel4, line 141: NB: See https://github.com/Jellyfish-AI/jf_agent/pull/219/files#r1182716687

jf_agent/data_manifests/git/generator.py
```diff
@@ -22,12 +21,41 @@ class UnsupportedGitProvider(Exception):
...
+            instance_slug = git_config.git_instance_slug
+            instance_info = endpoint_git_instances_info.get(instance_slug)
+            instance_creds = creds.git_instance_to_creds.get(instance_slug)
+        else:
+            # support legacy single-git support, which assumes only one available git instance
+            instance_info = list(endpoint_git_instances_info.values())[0]
+            instance_creds = list(creds.git_instance_to_creds.values())[0]
+
         try:
+            instance_slug: str = instance_info['slug']
```
jruel4, line 46: Q: Why do we retrieve the instance slug here? In the case of legacy single-git, this makes sense, but for multi-get we'd be re-retrieving the slug and replacing the one we get from our git_config.

I assume this won't actually cause any issue (since the one we get back from the endpoint_info will be the same), just wondering about the thinking behind this.

jruel4 (changes requested):
A few changes requested, but overall looks good.

jf_agent/data_manifests/git/generator.py
gavinpitt-jf: The main reason is that I was mirroring the existing agent logic as much as possible here. The secondary reason is that I think it'd better to have the 'truth' values come from the jellyfish app, rather than the the config.yaml. There could be one day where we need to overwrite the slugs in the yaml for some reason (very unlikely, but you never know), so I think it makes sense to keep it this way even if it is a little redundant

That being said, I should really dedup this logic between the existing git ingestion logic... I'll do that now

jruel4 (approved):
Changes look good, approved. ✅

PR #229, opened 2023-06-30
[OJ-26479] Log Error on Sprints 400

jruel4:
### Description
Log an error whenever we get a 400 back from the Jira sprints endpoint, rather than bombing out.


jobelenus (approved):

fialovy (approved):
Makes sense from my understanding 👍🏼

lukasbrower (approved):

PR #241, opened 2023-08-08
OJ-27617-refactor-github-ingest-to-use-gql

gavinpitt-jf:
Introduce a new Github Adapter that leverages Graphql to greatly increase speed of data ingestion. The biggest gains here will be realized by speeding up our pull request ingesiton

See the related Jira Ticket for a more in-depth analysis of this change

### DIFF
--- a/jf_agent/data_manifests/git/adapters/github.py
+++ b/jf_agent/data_manifests/git/adapters/github.py
@@ -105,7 +105,7 @@ class GithubManifestGenerator(ManifestAdapter):
         """
         path_to_page_info = 'data.organization.repositories'
         for result in self._page_results(
-            query_body=query_body, path_to_page_info=path_to_page_info,
+            query_body=query_body, path_to_page_info=path_to_page_info
         ):
             for repo in result['data']['organization']['repositories']['repos']:
                 yield GitRepoManifest(
@@ -152,7 +152,7 @@ class GithubManifestGenerator(ManifestAdapter):
 
         path_to_page_info = 'data.organization.users'
         for result in self._page_results(
-            query_body=query_body, path_to_page_info=path_to_page_info,
+            query_body=query_body, path_to_page_info=path_to_page_info
         ):
             for user in result['data']['organization']['users']['user_details']:
                 yield GitUserManifest(
@@ -191,7 +191,7 @@ class GithubManifestGenerator(ManifestAdapter):
 
         path_to_page_info = 'data.organization.repository.branches_query'
         for result in self._page_results(
-            query_body=query_body, path_to_page_info=path_to_page_info,
+            query_body=query_body, path_to_page_info=path_to_page_info
         ):
             for branch in result['data']['organization']['repository']['branches_query'][
                 'branches'
@@ -234,7 +234,7 @@ class GithubManifestGenerator(ManifestAdapter):
 
         path_to_page_info = 'data.organization.repository.prs_query'
         for result in self._page_results(
-            query_body=query_body, path_to_page_info=path_to_page_info,
+            query_body=query_body, path_to_page_info=path_to_page_info
         ):
             for pr in result['data']['organization']['repository']['prs_query']['prs']:
                 yield GitPullRequestManifest(
--- a/jf_agent/data_manifests/git/adapters/github.py
+++ b/jf_agent/data_manifests/git/adapters/github.py
@@ -15,7 +15,6 @@ from jf_agent.git.github_gql_utils import (
     get_raw_result,
     page_results,
 )
-from jf_agent.session import retry_session
 
 
 # TODO: Expand or generalize this to work with things other than github (BBCloud, Gitlab, etc)
@@ -40,7 +39,7 @@ class GithubManifestGenerator(ManifestAdapter):
         self.session = get_github_gql_session(token=token, verify=verify)
 
     def get_users_count(self) -> int:
-        body = f"""{{
+        query_body = f"""{{
             organization(login: "{self.org}"){{
                     users: membersWithRole {{
                         totalCount
@@ -49,9 +48,9 @@ class GithubManifestGenerator(ManifestAdapter):
             }}
         """
         # TODO: Maybe serialize the return results so that we don't have to do this crazy nested grabbing?
-        return get_raw_result(query_body=body, base_url=self.base_url, session=self.session)[
-            'data'
-        ]['organization']['users']['totalCount']
+        return self._get_raw_result(query_body=query_body)['data']['organization']['users'][
+            'totalCount'
+        ]
 
     def get_repos_count(self) -> int:
         query_body = f"""{{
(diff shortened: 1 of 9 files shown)


jf_agent/data_manifests/git/adapters/github.py
```diff
@@ -36,18 +33,10 @@ def __init__(
         # Session fields
         self.token = token
 
-        if base_url and 'api/v3' in base_url:
-            # Github server clients provide an API with a trailing '/api/v3'
-            # replace this with the graphql endpoint
-            self.base_url = base_url.replace('api/v3', 'api/graphql')
-        else:
-            self.base_url = 'https://api.github.com/graphql'
+        # Translate base URL if we are using cloud or server
```
CaptainKanuk, line 36: This looks like it's totally swapping over to the new method, are we doing that or feature flagging?

jf_agent/git/github_gql_adapter.py
```diff
@@ -0,0 +1,452 @@
...
+                    server_git_instance_info, nrm_repo.project.login, nrm_repo.id, 'commits'
+                )
+
+                try:
+                    for branch_name in get_branches_for_normalized_repo(
+                        nrm_repo, included_branches
+                    ):
+                        for j, api_commit in enumerate(
+                            tqdm(
+                                self.client.get_commits(
```
CaptainKanuk, line 146: The control flow here will fail to get the entire branch if we error out for any commit on the branch. Is that our intended behavior? It may be more robust to do

```
for branches
    try get commits
    for commits
        ...

```

jf_agent/git/github_gql_client.py
```diff
@@ -0,0 +1,466 @@
...
+                }}
+            }}
+        }}
+        """
+        for page in self._page_results(
+            query_body=query_body, path_to_page_info='data.organization.repoQuery'
+        ):
+            for api_repo in page['data']['organization']['repoQuery']['repos']:
+                # Skip over excluded or ignore non-included
+                if not all(filt(api_repo) for filt in repo_filters):
```
CaptainKanuk, line 156: We might waste time here. Is there a way we can do this by generating a set of include and exclude and checking set membership here instead?

jf_agent/git/github_gql_client.py
```diff
@@ -0,0 +1,466 @@
...
+                    continue
+                else:
+                    # Potentially process more branches
+                    if api_repo['branchQuery']['pageInfo']['hasNextPage']:
```
CaptainKanuk, line 160: I can see this logic getting a bit confusing in the future. Would be better to separate out our processing of branches into a separate flow? In ideal state we'd be submitting branches to a queue to process when we learn about them. I don't think we should change anything now but starting a discussion.

CaptainKanuk (commented):
Overall looks good - a few questions.

jf_agent/data_manifests/git/adapters/github.py
gavinpitt-jf: This is a little confusing, but this is the `data_manifests` gql adapter and not the one that the actual data processing uses. This has always used GQL so we're not hiding it behind the feature flag at all

gavinpitt-jf: There should be an effort to join this with the other `github_gql` adapter, but I decided to leave the seperately as we roll out the new feature. Ideally we'd consolidate them when we move things to the `jf_ingest` repository, or sometime before that

jf_agent/git/github_gql_adapter.py
gavinpitt-jf: Oh good call, I'll rework that

jf_agent/git/github_gql_client.py
gavinpitt-jf: Not sure exactly what you mean, I think that's what this is doing (I mostly copy-pasted this from the existing code)

I'm not actually super familiar with this all filter in python, but isn't it doing a simple 'is this string in the set' or not check for each repo name it finds?

```

        filters = []
        if self.config.git_include_repos:
            filters.append(
                lambda r: r['name'].lower()
                in set([r.lower() for r in self.config.git_include_repos])
            )
        if self.config.git_exclude_repos:
            filters.append(
                lambda r: r['name'].lower()
                not in set([r.lower() for r in self.config.git_exclude_repos])
            )
```

jf_agent/git/github_gql_client.py
gavinpitt-jf: Yeah this code assumes that for most repos that we process there are actually not more 50 branches on it (this repo only has 18, for example). It is a little confusing, but I think it makes sense to do this way. Branches are super simple data objects (NormalizedBranch only has a `name` and `sha` field) and the `NormalizedBranch` class is only used as a subclass on `NormalizedRepository`. Intertwining them here makes sense to me, and it gives us a potential speed boost if we guess correctly that there less than 50 branches in a repo.

jf_agent/git/github_gql_adapter.py
gavinpitt-jf: I reworked it to match how it was functioning before under `jf_agent/git/github.py`. Now if we run into any issues we will still potentially will lose commits but we won't exist early on processing branches. Here is the snippet from the 'legacy' way. The get commits call returns a generator, so it's a little awkward to wrap it in a standalone try-catch block like I think you suggested. I also don't want to put it in the client itself, because I want users of the client to know when we fail to get a commit

```
            for branch in branches_to_process:
                try:
                    for j, commit in enumerate(
                        tqdm(
                            client.get_commits(
                                repo['full_name'], branch, since=pull_since, until=None
                            ),
                            desc=f'downloading commits on branch {branch} for {repo["name"]}',
                            unit='commits',
                        ),
                        start=1,
                    ):
                    ...
```

jf_agent/data_manifests/git/adapters/github.py
CaptainKanuk: Ah yeah makes sense. missed that.

jf_agent/git/github_gql_adapter.py
CaptainKanuk: Got it - I think preserving old behavior is the right call here to avoid changing too many things at once. In the future we probably will want to separately make a call on whether the overall control flow, and when we choose to try/except vs. raise, is optimal. My suspicion is that it could be a lot better.

CaptainKanuk (approved):

jf_agent/git/github_gql_client.py
CaptainKanuk: Ah - I misunderstood the previous code. I thought it was creating an array of lambdas, but it only creates 2 and each is a set membership test. That makes sense. The code is slightly confusing because the `all` function makes it feel like we're iterating over a variety of filters when in reality it's only 2, but I don't think it's worth changing now.

PR #253, opened 2023-08-17
OJ-28629-disable-top-level-comments-from-getting-pulled

gavinpitt-jf:
Disable 'Top Level' comments on reviews, akak 'IssueComments' vs 'PullRequestReviewComments' in the GQL schema

This is inline with how we are doing comments in the 'legacy' github ingest, and better reflects how comments are tracked by our metrics

### DIFF
--- a/jf_agent/git/__init__.py
+++ b/jf_agent/git/__init__.py
@@ -290,6 +290,7 @@ def load_and_dump_git(
     outdir: str,
     compress_output_files: bool,
     git_connection,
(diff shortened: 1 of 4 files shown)


thingscouldbeworse (changes requested):
I think we should do a quick wrap of `if jf_options.get('get_all_issue_comments'):` and feature flag this instead of just pulling it out

thingscouldbeworse (approved):
👍

PR #254, opened 2023-08-17
Oj 28632 reduce number of api calls for prs

gavinpitt-jf:
This pull requests gets rid of the function that is used to determine the page size when pulling repos. This page size function was initially introduced to help us quickly see if there are no PRs that we need to pull, but it has since been replace by the `reop_has_quiescent_prs_lookup` look up table. Getting rid of this superfluous page size checking will help reduce our total number of calls and fly lower under the GQL API rate limiter


thingscouldbeworse (approved):
good stuff

PR #262, opened 2023-08-23
OJ-27738-upgrade-python

gavinpitt-jf:
Upgrade python from 3.9 to 3.10.12

This is good for security purposes, and will allow us to use the developing `jf_ingest` module which requires `>=3.10`


thingscouldbeworse (approved):

pyproject.toml
```diff
@@ -68,7 +68,7 @@ dependencies = [
     "click~=8.0.4",
     "requests>=2.31.0",
 ]
-requires-python = ">=3.9"
+requires-python = ">=3.10"
```
ametzger, line 71: nit: should update the `tools.black` block in this file to set the target version to py310

ametzger (approved):

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.black]
 line-length = 100
 skip-string-normalization = true
-target_version = ['py37', 'py38']
+target_version = ['py310']
```
gavinpitt-jf, line 4: @ametzger is this what you meant/the correct approach? I got rid of the py37/py38 values and set it to be py310 only

ametzger: Yup, perfect, thanks!

PR #263, opened 2023-08-23
OJ-28807-at-py-test-via-pdm

gavinpitt-jf:
Add in the `pytest` library to PDM. Also clean up a few problems detected when testing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,6 @@ dependencies = [
     "black~=19.10b0",
     "click~=8.0.4",
     "requests>=2.31.0",
-    "pytest>=7.4.0",
 ]
 requires-python = ">=3.10"
 readme = "README.md"
@@ -82,6 +81,7 @@ name = "pypi"
 [tool.pdm.dev-dependencies]
 dev = [
     "requests-mock",
+    "pytest>=7.4.0",
 ]
 
 [build-system]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.black]
 line-length = 100
 skip-string-normalization = true
-target_version = ['py310']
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.black]
 line-length = 100
 skip-string-normalization = true
-target_version = ['py310']
+target_version = ['py37', 'py38']
```
gavinpitt-jf, line 4: @ametzger as noted in slack, I'm rolling this back. I think we need to update `black` before we can use it with py310?

tests/test_jira_download.py
```diff
@@ -90,18 +100,24 @@ def one_bad_issue_callback(self, request, context):
     @classmethod
     def setUpClass(cls):
         cls.jira_connection = get_connection()
-        with open("test_data/jira/test_issues_response.json", "r") as issues_file:
+        with open("tests/test_data/jira/test_issues_response.json", "r") as issues_file:
```
gavinpitt-jf, line 103: @thingscouldbeworse - most of the files changes here are formatting, but this was an intentional change. For some reason this base file path was missing in this test. It was causing errors for me when run on the command line

pyproject.toml
```diff
@@ -67,6 +67,7 @@ dependencies = [
     "black~=19.10b0",
     "click~=8.0.4",
     "requests>=2.31.0",
+    "pytest>=7.4.0",
```
acicchetti-jf, line 70: Should this be a dev dependency?

gavinpitt-jf: Yes

thingscouldbeworse (approved):

acicchetti-jf (approved):

PR #285, opened 2023-10-25
OJ-30360-fix-page-batching-bad-logic

gavinpitt-jf:
## Description

Fix some bad batching logic when fetching a large amount of Jira issues. In a function that batches pages, we were returning with a `return` statement instead of using a yield statement.

## Testing

To Test this I spoofed our functions to pull a years worth of orthogonal-networks data, along with some logic changes to make our batches MUCH smaller to throttle the batching logic. I was able to fetch all issue data with the 'server safety' throttle batching things in groups of 7

### DIFF
--- a/jf_agent/jf_jira/jira_download.py
+++ b/jf_agent/jf_jira/jira_download.py
@@ -6,7 +6,7 @@ import random
 import re
 import traceback
 import time
-from typing import Dict
+from typing import Dict, Generator, Tuple
 
 from dateutil import parser
 from jira.exceptions import JIRAError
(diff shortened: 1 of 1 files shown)


jf_agent/jf_jira/jira_download.py
```diff
@@ -755,17 +756,21 @@ def _download_jira_issues_page(
...
                         )
-                        return [], 0
+                        yield ([], 0)
+                        break
                     else:
                         get_changelog = False
                         batch_size = 100
                         start_at += 1
 
+    return [], 0
```
thingscouldbeworse, line 772: This is the end case where we never yielded anything right? And also just the final return after yielding everything else?

thingscouldbeworse (approved):
Looks good as I understand it, if it was tested then ship

PR #291, opened 2023-11-01
slice down issue ids to fetch per segment

ametzger:


thingscouldbeworse (approved):
tests pass, run locally, lgtm 👍

PR #293, opened 2023-11-02
OJ-30558: logging params on all and extra http debug if set with run flag

thingscouldbeworse:


jf_agent/main.py
```diff
@@ -107,6 +107,9 @@ def main():
     parser.add_argument(
         '-u', '--until', nargs='?', default=None, help='DEPRECATED -- has no effect'
     )
+    parser.add_argument(
+        '-dr', '--debug-request', type=bool, default=False, help='turn on verbose http logging on errors'
+    )
```
ametzger, line 112: nit: should use `store_true` here so we can just provide the flag rather than having to do `--debug-request True`

```suggestion
    parser.add_argument(
        '-dr',
        '--debug-request',
        action='store_true',
        help='turn on verbose http logging on errors',
    )
```

PR #296, opened 2023-11-17
[OJ-30980] Move validate run mode to use jf_ingest validation methods

nathanieltalbot:
A few medium-sized changes to agent code:
- Cuts over the validation run mode to use the validate_function that has been moved into jf_ingest.
- Adds a helper function, `get_jira_ingest_config`, that will take in the agent config and credentials and output a JiraIngestionConfig as defined in jf_ingest.
- Moves around some logic, including the get_company_info call, which is now encapsulated in it's own function.
- Fixes an issue with validate_memory to use the config file's outdir rather than a hardcoded one.

This will make it easier for us to use the jf_ingest validation code and other parts of jf_ingest Jira code, and will help share this logic to aid debugging.

### DIFF
--- a/jf_agent/config_file_reader.py
+++ b/jf_agent/config_file_reader.py
@@ -7,8 +7,13 @@ from typing import List
 import urllib3
 import yaml
 
-from jf_agent import VALID_RUN_MODES, BadConfigException
+from jf_agent import VALID_RUN_MODES
+from jf_agent.exception import BadConfigException
 from jf_ingest import logging_helper
+from jf_ingest.jf_jira import JiraIngestionConfig
+from jf_ingest.jf_jira.auth import JiraAuthConfig
+
+from jf_agent.util import get_company_info
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -68,7 +68,7 @@ dependencies = [
     "click~=8.0.4",
     "requests>=2.31.0",
     "python-dotenv>=1.0.0",
-    "jf-ingest==0.0.10",
+    "jf-ingest>=0.0.24",
```
gavinpitt-jf, line 71: I'd prefer if this was locked into 0.0.24, rather than 0.0.24 and above. Ingest is getting rolled out bit by bit and I'd prefer that we lock Agent and D.I. into confirmed working versions, rather than somebody risk updating JF ingest and not making necessary changes on the agent side to match it

nathanieltalbot: Fixed!

jf_agent/config_file_reader.py
```diff
@@ -291,6 +296,70 @@ def _get_git_config_from_yaml(yaml_config) -> List[GitConfig]:
...
+        else:
+            auth_config = JiraAuthConfig(
+                company_slug=company_slug,
+                url=config.jira_url,
+                personal_access_token=creds.jira_bearer_token,
+                bypass_ssl_verification=config.skip_ssl_verification
+
+            )
+    else:
+        raise BadConfigException("Cannot find jira credentials!")
```
acicchetti-jf, line 330: This section is smelly - could it be phrased differently?
```text
if not config.jira_url:
	Bad Jira URL

if jira_username and password:
	auth_config=…
elif jira_bearer_token:
	auth_config=…
else:
	Bad Jira Credentials
```

acicchetti-jf (approved):
Nit is there, would prefer it gets changed, but won't block on it

jf_agent/config_file_reader.py
nathanieltalbot: Good call, fixed this

acicchetti-jf (approved):
Awesome, TY. G2G

PR #297, opened 2023-11-29
[OJ-31074] Uploading report of connectivity health during validation run mode

nathanieltalbot:
Uploads the validation/healthcheck report during the validation run mode. Also bumps the jf_ingest version, and adds a config variable to set whether or not to disable the upload.

This does not upload during a standard agent run -- that will come in another PR.

### DIFF
--- a/jf_agent/data_manifests/git/generator.py
+++ b/jf_agent/data_manifests/git/generator.py
@@ -18,6 +18,13 @@ class UnsupportedGitProvider(Exception):
     pass
 
 
+def get_instance_slug(git_config, is_multi_git_config: bool, endpoint_git_instances_info) -> str:
+    if is_multi_git_config:
+        return git_config.git_instance_slug
+    else:
+        instance_info = list(endpoint_git_instances_info.values())[0]
+        return instance_info['slug']
+
 def create_manifests(
     company_slug: str,
     creds: dict,
@@ -31,6 +38,8 @@ def create_manifests(
 
     # Iterate over each git config within the config yaml
     for git_config in git_configs:
+        instance_slug = get_instance_slug(git_config, is_multi_git_config, endpoint_git_instances_info)
+
         if is_multi_git_config:
             instance_slug = git_config.git_instance_slug
             instance_info = endpoint_git_instances_info.get(instance_slug)
--- a/jf_agent/data_manifests/git/generator.py
+++ b/jf_agent/data_manifests/git/generator.py
@@ -18,12 +18,16 @@ class UnsupportedGitProvider(Exception):
     pass
 
 
-def get_instance_slug(git_config, is_multi_git_config: bool, endpoint_git_instances_info) -> str:
-    if is_multi_git_config:
+def get_instance_slug(git_config: GitConfig, endpoint_git_instances_info: dict) -> str:
+    """
+    Returns the GitConfig's instance slug.
+    """
+    if git_config.git_instance_slug:
         return git_config.git_instance_slug
     else:
         instance_info = list(endpoint_git_instances_info.values())[0]
         return instance_info['slug']
+
 
 def create_manifests(
     company_slug: str,
@@ -38,17 +42,15 @@ def create_manifests(
 
     # Iterate over each git config within the config yaml
     for git_config in git_configs:
-        instance_slug = get_instance_slug(git_config, is_multi_git_config, endpoint_git_instances_info)
+        instance_slug = get_instance_slug(git_config, endpoint_git_instances_info)
 
         if is_multi_git_config:
-            instance_slug = git_config.git_instance_slug
             instance_info = endpoint_git_instances_info.get(instance_slug)
             instance_creds = creds.git_instance_to_creds.get(instance_slug)
         else:
             # support legacy single-git support, which assumes only one available git instance
             instance_info = list(endpoint_git_instances_info.values())[0]
             instance_creds = list(creds.git_instance_to_creds.values())[0]
-            instance_slug: str = instance_info['slug']
 
         try:
             # If the git config doesn't have a git instance, do not generate a manifest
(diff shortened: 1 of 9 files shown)


jf_agent/data_manifests/git/generator.py
```diff
@@ -18,6 +18,13 @@ class UnsupportedGitProvider(Exception):
     pass
 
 
+def get_instance_slug(git_config, is_multi_git_config: bool, endpoint_git_instances_info) -> str:
```
CaptainKanuk, line 21: I don't like the pattern that this function has to be parameterized with `is_multi_git_config`. Could we instead have this function inspect the config and determine that itself?

CaptainKanuk: I view it as leading to later complexity and bugs to have part of the necessary functionality live in this function and part not.

jf_agent/validation.py
```diff
@@ -58,23 +67,32 @@ def full_validate(config, creds):
     # Check for Git configs
     if config.git_configs:
         try:
-            git_success = validate_git(config, creds)
+            git_healthcheck_result = validate_git(config, creds, jellyfish_endpoint_info.git_instance_info)
```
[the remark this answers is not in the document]
nathanieltalbot: Can do, I can put a change into ingest with the name change as well

jf_agent/data_manifests/git/generator.py
nathanieltalbot: The logic for this would basically be:
```
if git_instance_slug in git_config:
    return git_config.git_instance_slug
else:
    # Get slug from API
 
```

Just to confirm, is an instance having a git instance slug always an indicator that it is a multi-git config?

jf_agent/validation.py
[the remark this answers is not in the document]
nathanieltalbot: Ingest PR to do this: https://github.com/Jellyfish-AI/jf_ingest/pull/38

jf_agent/data_manifests/git/generator.py
gavinpitt-jf: @nathanieltalbot I don't know if it's an indicator that it's specifically a mulit-git config, but it is impossible to have a multi git config WITHOUT a slug for each: https://github.com/Jellyfish-AI/jf_agent/blob/3f37c6025ecd34ffc78026504dd2eb5b8e25083d/jf_agent/config_file_reader.py#L386

That logic you linked in the above comment should hold, because a single git instance MIGHT have a slug provided, but if it doesn't we can fall back on the API. If we're in a multi config, than we know it WILL have a slug

gavinpitt-jf: Agree with delany here, it'd be nice to decouple the `is_multi_git_config` logic here. Also it would be nice to add some type casting: `(git_config: GitConfig, is_multi_git_config: bool, endpoint_git_instances_info: dict)`

jf_agent/main.py
```diff
@@ -128,16 +128,16 @@ def main():
 
     success = True
 
+    jellyfish_endpoint_info = obtain_jellyfish_endpoint_info(config, creds)
```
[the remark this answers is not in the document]
gavinpitt-jf: I think I might be being over paranoid, and we might already do checks for that. Consider this a NIT

jf_agent/validation.py
```diff
@@ -3,9 +3,13 @@
 import psutil
 import shutil
 
+import requests
+
 from jf_agent.config_file_reader import get_jira_ingest_config
+from jf_agent.data_manifests.git.generator import get_instance_slug
 from jf_agent.git import get_git_client, get_nested_repos_from_git, GithubGqlClient
-from jf_ingest.validation import validate_jira
+from jf_ingest.validation import validate_jira, GitHealthCheckResult, JiraHealthCheckResult, IngestionHealthCheckResult, IngestionType
```
gavinpitt-jf, line 11: 🔥

jf_agent/validation.py
```diff
@@ -58,23 +67,32 @@ def full_validate(config, creds):
...
-
     else:
         logger.info("\nNo Git configs provided, skipping Git validation...")
 
     # Finally, display memory usage statistics.
     validate_memory(config)
 
+    healthcheck_result: IngestionHealthCheckResult = IngestionHealthCheckResult(ingestion_type=IngestionType.AGENT,
+                                                                                git_healthcheck_result=git_healthcheck_result,
+                                                                                jira_healthcheck_result=jira_healthcheck_result)
```
gavinpitt-jf, line 86: Do we have black set up as a git hook? This is obviously a formatting NIT, I'm just surprised black left it like this

nathanieltalbot: For agent I don't think so? Sometimes PyCharm does/recommends it's own formatting though.

jf_agent/validation.py
```diff
@@ -200,3 +245,19 @@ def validate_memory(config):
     except Exception as e:
         print(f"  ERROR: Could not obtain memory and/or disk usage information. {e}")
         return False
+
+
+def upload_to_s3(jellyfish_api_base: str, jellyfish_api_token: str, healthcheck_result: IngestionHealthCheckResult) -> None:
```
gavinpitt-jf, line 250: NIT: this technically isn't uploading to S3, it's uploading to jellyfish to then go to s3, so maybe `submit_health_check_to_jellyfish` would be a better name. That's pretty pedantic of me tho 😜

gavinpitt-jf (approved):
Looks good to me, my comments were mostly stylistic NITS

jf_agent/data_manifests/git/generator.py
nathanieltalbot: Just changed this, thanks!

jf_agent/validation.py
nathanieltalbot: Fixed!

PR #299, opened 2023-12-11
fix: Implement proper rate limit handling in github_client

kxs-fchamseddine:
# The problem

We've been running into intermittent failures when running the Jellyfish agent in "download_only" mode which turned out to be caused by secondary rate limits: though `github_gql_client` properly handles both primary and secondary rate limits, `github_client` only properly handles primary rate limits. When we were hitting the secondary rate limit, a single retry would occur (without waiting) before an exception is thrown since it's not a primary rate limit:
`
[3061] Failed to download github data:
403 Client Error: Forbidden for url: https://api.github.com/users/Kinaxis
`

# The solution

Rather than the currently implemented hack, we can take a similar approach to what's done in `github_gql_client`:
The documentation states that the presence of `retry-after` in the response headers indicates that a secondary rate limit was hit, and we can thus utilize this information with the existing retry mechanism to perform a reasonable sleep before retrying a select number of times.

The primary rate limit will be prioritized, as is recommended by the GitHub API documentation.

Exceptions will still be raised for non-rate limit related errors.

We have confirmed that this fix has greatly improved stability over the last week during our forks' automated builds.

# Manual testing

## Given

A response header containing a `X-RateLimit-Remaining` value of 0 (indicating a primary rate limit was hit)

## When

We run the agent in "download_only" mode

## Then

A retry dictated by `X-RateLimit-Reset` is conducted, as is indicated by the 3091 warning log:
`
[3091] Github rate limit exceeded.  Trying again in 0:47:56.194486...
`

## Given

A response header containing a `retry-after` non-zero value (indicating a secondary rate limit was hit)

## When

We run the agent in "download_only" mode

## Then

A retry dictated by `retry-after` is conducted, as is indicated by the 3091 warning log:
`
[3091] Github rate limit exceeded.  Trying again in 0:01:00...
`

### DIFF
--- a/jf_agent/git/github_client.py
+++ b/jf_agent/git/github_client.py
@@ -121,21 +121,29 @@ class GithubClient:
         for i in range(1, max_retries + 1):
             try:
                 result = self.session.get(url)
-
-                # HACK: This appears to happen after we have been
-                # rate-limited when hitting certain URLs, there is
-                # likely a more elegant way to solve this but it takes
-                # about an hour to test each time and it works.
-                if result.status_code == 403:
-                    result = self.session.get(url)
-
                 result.raise_for_status()
                 return result
             except requests.exceptions.HTTPError as e:
-                remaining_ratelimit = e.response.headers.get('X-RateLimit-Remaining')
-                ratelimit_reset = e.response.headers.get('X-RateLimit-Reset')
+                reset_wait_in_seconds = 0
+                # Primary rate limits should take precedence over secondary rate limits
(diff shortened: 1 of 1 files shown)


kxs-fchamseddine:
We've opened a Jellyfish ticket to get someone to review this PR. This was when we were informed that they are able to flick a switch on their side to enforce the usage of the GQL client when querying GitHub which already properly handles all rate limits.

After synchronizing, this change was conducted. There is thus no longer a need for these changes and the PR can be closed.

PR #308, opened 2023-12-21
[OJ-31903] Upload on Failure or Timeout

nathanieltalbot:
A "dead man's switch" of sorts, to allow us to upload logs in the case of _any_ failure, at least, hopefully.

Basically:
- In rollback_on_fail.sh, if we find we've failed, we kick off `main.py` _again_ with the `-f` flag, indicating `from failure`
- This will execute a fun special mode that doesn't download, but just sets the output directory to whatever we find as the last directory in the outdir.
- There's a catch-22 here that I'm trying to patch wherein we don't actually get the logs from _this_ send-only run of the agent, but working on fixing that by adding the old logfile to the curr directory.

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -10,18 +10,18 @@ from collections import namedtuple
 from glob import glob
 from pathlib import Path
 import sys
-from time import sleep
 
 import requests
 import json
+
 
 from jf_agent import (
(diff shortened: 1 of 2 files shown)


thingscouldbeworse (approved):
Approach makes sense

PR #312, opened 2024-01-05
OJ-32077-make-jira-fail-a-soft-fail

gavinpitt-jf:
Previously, before JF Ingest, if JIRA failed it would be a 'soft' failure and move on to attempt to download GIT data as well as upload logging and diagnostic information. With the rollout of JF Ingest, however, a bug was introduced that made JIRA failures a hard failure. This PR seeks to address that and make the JIRA failure soft again

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -1,5 +1,6 @@
 import argparse
 from concurrent.futures import ThreadPoolExecutor
+import traceback
 import dotenv
 import gzip
 import logging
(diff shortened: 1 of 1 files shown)


thingscouldbeworse (commented):
We need to report back `success` to the status file don't we?

gavinpitt-jf:
> We need to report back `success` to the status file don't we?

Yup, that's what this `append` statement is doing

```
            success_status = 'success' if success else 'failed'
            download_data_status.append({'type': 'Jira', 'status': success_status})
```

thingscouldbeworse (approved):

PR #314, opened 2024-01-12
OJ-32108-agent-side-changes-for-memory-foot-print-reduction

gavinpitt-jf:
Bump JF Ingest version and de-risk generating JF Ingest config. JF Ingest will be controlled by a separate agent only config flag on the Jellyfish side, so this release will be restricted to only clients that have the config flag enabled

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -200,9 +200,6 @@ def main():
             diagnostics.capture_run_args(
                 args.mode, args.config_file, config.outdir, args.prev_output_dir
             )
-            ingest_config = get_ingest_config(
-                config=config, creds=creds, endpoint_jira_info=jellyfish_endpoint_info.jira_info
-            )
             try:
                 generate_manifests(
                     config=config, creds=creds, jellyfish_endpoint_info=jellyfish_endpoint_info
@@ -226,7 +223,6 @@ def main():
                 download_data_status = download_data(
                     config,
                     creds,
-                    ingest_config,
                     jellyfish_endpoint_info.jira_info,
                     jellyfish_endpoint_info.git_instance_info,
                     jellyfish_endpoint_info.jf_options,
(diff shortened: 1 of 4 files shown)


jf_agent/main.py
```diff
@@ -200,9 +200,6 @@ def main():
             diagnostics.capture_run_args(
                 args.mode, args.config_file, config.outdir, args.prev_output_dir
             )
-            ingest_config = get_ingest_config(
-                config=config, creds=creds, endpoint_jira_info=jellyfish_endpoint_info.jira_info
-            )
```
gavinpitt-jf, line 205: Move ingest config out of here, where we generate it every time. Move it to the logic block that only triggers if the config flag is set

jf_agent/main.py
```diff
@@ -575,31 +571,33 @@ def distribute_repos_between_workers(git_configs, metadata_by_project):
 
 @diagnostics.capture_timing()
 @logging_helper.log_entry_exit(logger)
-def download_data(
-    config, creds, ingest_config, endpoint_jira_info, endpoint_git_instances_info, jf_options
-):
+def download_data(config, creds, endpoint_jira_info, endpoint_git_instances_info, jf_options):
```
gavinpitt-jf, line 574: remove ingest_config as an argument

jf_agent/main.py
```diff
@@ -575,31 +571,33 @@ def distribute_repos_between_workers(git_configs, metadata_by_project):
...
                 )
                 logger.debug(traceback.format_exc())
-                success = False
+                logger.info(f'Rolling back and using load_and_dump_jira to ingest data')
 
-            success_status = 'success' if success else 'failed'
-            download_data_status.append({'type': 'Jira', 'status': success_status})
+                download_data_status.append(
+                    load_and_dump_jira(config, endpoint_jira_info, jira_connection)
+                )
```
gavinpitt-jf, line 600: If JF Ingest does fail, attempt to run the legacy way. This will derisk the initial launch of jf ingest

jruel4 (approved):

PR #317, opened 2024-02-05
OJ 32624 OJ-32625  add logging queue handler

breid1313:
Add a log queue listener and handler to the logging config that will periodically post log messages to Jellyfish to reduce the latency in receiving agent logs.

### DIFF
--- a/jf_agent/agent_logging.py
+++ b/jf_agent/agent_logging.py
@@ -1,7 +1,11 @@
 import logging
+from logging import LogRecord
 import os
 import sys
-import urllib3
+from logging.handlers import QueueHandler, QueueListener
+from queue import Queue
+from dataclasses import dataclass
+from typing import List
 
 '''
 Guidance on logging/printing in the agent:
@@ -19,6 +23,7 @@ e.g., function names, iteration counts
 '''
 
 LOG_FILE_NAME = 'jf_agent.log'
+
 
 # For styling in log files, I think it's best to always use new lines even when we use
 # the special character to ignore them. Leverage always_use_newlines for this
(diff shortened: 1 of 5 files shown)


jf_agent/main.py
```diff
@@ -233,15 +249,13 @@ def main():
...
             diagnostics.capture_outdir_size(config.outdir)
-            diagnostics.close_file()
-
-            # Kills the sys_diag_collector thread
-            sys_diag_done_event.set()
-            sys_diag_collector.join()
 
         except Exception as err:
             logger.error(f"Encountered error during agent run! {err}")
+            agent_logging.close_out(logging_config)
```
jruel4, line 255: Q: Is this something that should live in the "finally" block?

jf_agent/util.py
```diff
@@ -84,5 +88,30 @@ def upload_file(filename, path_to_obj, signed_url, config_outdir, local=False):
...
+        "urllib3.util.retry",
+        "urllib3.util",
+        "urllib3",
+        "urllib3.connection",
+        "urllib3.response",
+        "urllib3.connectionpool",
+        "urllib3.poolmanager",
+        "requests",
+    ]
+    for name in target_loggers:
```
jruel4, line 110: Q: Slightly concerned that this will introduce a race condition where we occasionally miss log messages while we're posting to Jellyfish. Maybe we can do a non-blocking lock check on the `post_logs_to_jellyfish` function instead?

jf_agent/agent_logging.py
```diff
@@ -60,22 +71,105 @@ def emit(self, record) -> None:
...
+        if resp.ok:
+            self.messages_to_send = []
+            self.last_message_send_time = now
+            if self.create_stream and resp.json()['stream_created']:
+                self.create_stream = False
+
+    def handle(self, record: Union[logging.LogRecord, int]) -> None:
+        now = datetime.now()
+
+        if record is not self._sentinel:
```
jruel4, line 115: Nit: In this instance, better practice to use `==`. Generally should reserve `is` for actually checking that two objects have the same identity (e.g. are the same object), rather than checking equality.

See here for a good explanation on this: https://stackoverflow.com/a/25758019

jf_agent/agent_logging.py
```diff
@@ -60,22 +71,105 @@ def emit(self, record) -> None:
...
+                            record.asctime + '000', '%Y-%m-%d %H:%M:%S,%f'
+                        ).timestamp()
+                        * 1000
+                    ),
+                    'initiated_at': self.initiated_at,
+                }
+            )
+
+        if (
+            record is self._sentinel
```
jruel4, line 131: Same as https://github.com/Jellyfish-AI/jf_agent/pull/317/files#r1479175955

jf_agent/util.py
breid1313: I agree this felt sub-optimal when writing. I was interested in getting a POC going, but have now re-factored to use a lower level http connection library that does not log 🚀

jf_agent/main.py
```diff
@@ -233,15 +249,13 @@ def main():
                 )
 
             diagnostics.capture_outdir_size(config.outdir)
-            diagnostics.close_file()
-
-            # Kills the sys_diag_collector thread
-            sys_diag_done_event.set()
-            sys_diag_collector.join()
```
breid1313, line 240: we were doing this in multiple places that could/should just be handled by a `finally`

thingscouldbeworse (approved):

PR #319, opened 2024-02-08
sort the list of logs every time we add an item

breid1313:
ensure that the list of logs to post is always sorted by timestamp. I am choosing to do this each time that we add a record to the list rather than at post time in the event that there is a race condition that adds a record to the queue while we are preparing to post


breid1313:
whoops this should be done on the other end

PR #323, opened 2024-03-01
OJ-33445: fix roll back logic when jf ingest crashes

gavinpitt-jf:
Adjust logic to delete left over JF Ingest files when JF Ingest crashes and rolls back to not use JF Ingest

I tested this by incorrectly setting my GDPR_active flag to False in my config.yml, which recreates a specific problem one of our customers experienced

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -227,9 +227,7 @@ def main():
 
             if config.run_mode_is_print_apparently_missing_git_repos:
                 issues_to_scan = get_issues_to_scan_from_jellyfish(
-                    config,
-                    creds,
-                    args.for_print_missing_repos_issues_updated_within_last_x_months,
+                    config, creds, args.for_print_missing_repos_issues_updated_within_last_x_months,
                 )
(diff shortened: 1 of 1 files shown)


jf_agent/main.py
```diff
@@ -612,6 +605,21 @@ def download_data(config, creds, endpoint_jira_info, endpoint_git_instances_info
...
+                        )
+                        shutil.rmtree(jf_ingest_files)
+                        logger.debug(f'Successfully removed left over JF Ingest files')
+                    else:
+                        logger.debug(f'No JF Ingest files were detected at {jf_ingest_files}')
+                except Exception as e:
+                    # Extreme precaution, wrap the above file operations in a try catch block
+                    logger.debug(
+                        f'Error encountered when attempting to remove JF Ingest files: {e}'
+                    )
```
gavinpitt-jf, line 622: This is the only 'change'. Seems like my formatter went crazy on random stuff in this file 🤔

thingscouldbeworse (approved):

PR #325, opened 2024-03-14
Oj 33568  upload health test results during full run

breid1313:
- always run Jira validation via jf_ingest when in a pull mode
- run git validation if asked for, but not yet through jf_ingest and only when in validate mode
- bump jf_ingest so we can dump the dataclasses safely to json

taking a look at the dev bucket I can see the healthcheck file was uploaded along with the other json files.

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -178,7 +178,9 @@ def main():
 
     elif config.run_mode == 'validate':
         try:
-            full_validate(config, creds, jellyfish_endpoint_info, upload=not config.skip_healthcheck_upload)
+            full_validate(
+                config, creds, jellyfish_endpoint_info, upload=not config.skip_healthcheck_upload
+            )
         except Exception as err:
             logger.error(
                 f"Failed to run healthcheck validation due to exception, moving on. Exception: {err}"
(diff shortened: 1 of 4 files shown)


jf_agent/main.py
```diff
@@ -191,7 +179,15 @@ def main():
 
         logger.info(f"Mounted old output directory as {config.outdir}, will attempt to send.")
 
-    else:
+    elif not config.run_mode == 'send_only':
+        # Importantly, don't overwrite the already-existing diagnostics file
+        try:
+            full_validate(config, creds, jellyfish_endpoint_info)
```
nathanieltalbot, line 185: So FYI on this -- I disabled this back in December because it made the runtimes for larger companies like eBay really long (https://github.com/Jellyfish-AI/jf_agent/pull/310). The culprit I think was the nested repos and projects dict, which made an enormous amount of queries to the Github API to get all branches for all repos, which ran way too long: https://github.com/Jellyfish-AI/jf_agent/blob/master/jf_agent/validation.py#L178. Unsure if that function has been changed since, but this is why I disabled it

breid1313: thats a great point. I think what I will do is enable the jira side from jf_ingest (takes ~10-15 mins in the worst case) and skip git temporarily since we are migrating that over to jf_ingest as well. then at a later time I will add the git validation back in via jf_ingest

PR #326, opened 2024-03-18
OJ-33810: update jf ingest and tqdm dependencies

gavinpitt-jf:
## Description

Update dependencies for TQDM and JF Ingest, which we believe may be related to a strange string formatting error we're seeing for certain clients when they hit the BitbucketCloud RateLimiter


breid1313 (approved):
worth a go

PR #330, opened 2024-04-05
OJ-34303: remove jf ingest attempted roll back on exception handling

gavinpitt-jf:
JF Ingest (for Jira download) is stable enough now that we can remove the rollback logic on failure. The idea here is that if we are failing in JF Ingest here, than we'll likely fail in the rollback. Removing the rollback logic simplifies our code and fails the agent faster. It also simplifies our S3 upload, so we don't have to potentially check for BOTH file formats on a rollback event


acicchetti-jf (approved):
Seems good to me

PR #343, opened 2024-07-10
OJ-36885: always close down diagnostics file and thread

gavinpitt-jf:
## Description

There is a latent bug in the diagnostics monitoring thread. The logic flow to shut down the monitoring thread requires us to be in a mode where we are downloading/sending data. If the Agent is run in a `send_only` mode or `from_failure` mode, it will fail to close down the diagnostics thread and hang forever (well, technically it will get shutdown after 72 hours due to some time out logic we have in the Agent)

To resolve this issue, I moved the `diagnostics` cleanup logic (where we shut down the thread) to be part of a `try/finally` logic block so that it always gets run for ever run mode and even if we encounter an error

### DIFF
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -277,16 +277,17 @@ def main():
                 return False
 
     finally:
-        logger.info('Closing Diagnostics file')
-        # We need to close this before we send data
-        # Otherwise we'll send a .fuse_hidden file (temp file)
-        diagnostics.close_file()
 
         # Kills the sys_diag_collector thread.
         # We need to do this before exiting, otherwise we'll hang forever and never exit until timeout kills it.
         logger.info('Shutting down Systems Diagnostics Thread')
         sys_diag_done_event.set()
         sys_diag_collector.join()
+
+        logger.info('Closing Diagnostics file')
+        # We need to close this before we send data
+        # Otherwise we'll send a .fuse_hidden file (temp file)
+        diagnostics.close_file()
 
     success &= potentially_send_data(config, creds, successful=error_and_timeout_free)
 
--- a/jf_agent/main.py
+++ b/jf_agent/main.py
@@ -184,100 +184,109 @@ def main():
             creds.jellyfish_api_token,
         ),
     )
-    sys_diag_collector.start()
+    try:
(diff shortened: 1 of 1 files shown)


jf_agent/main.py
```diff
@@ -184,101 +184,109 @@ def main():
...
+            except Exception as err:
+                logger.error(f"Encountered error during agent run! {err}")
+                agent_logging.close_out(logging_config)
+                return False
+
+    finally:
+        logger.info('Closing Diagnostics file')
+        # We need to close this before we send data
+        # Otherwise we'll send a .fuse_hidden file (temp file)
+        diagnostics.close_file()
```
herbal-jelly, line 283: Should `diagnostics.close_file()` come after `sys_diag_done_event.join()`... there seems to be a possible race condition here.

herbal-jelly: Also, are we ok with `diagnostics.send_diagnostic_end_reading` potentially not getting called if we return early?  or, should that be part of the finally?

herbal-jelly (approved):
Just a couple questions, but otherwise looks good to me.

jf_agent/main.py
gavinpitt-jf: > Also, are we ok with `diagnostics.send_diagnostic_end_reading` potentially not getting called if we return early? or, should that be part of the finally?

This one is kind of a toss up for me. If we end becuase of an error and the `heartbeat` stops with no `stop` signal sent than we should be able to quickly tell that it crashed. I could also see sending the `stop` signal with a status attribute working pretty well too... Truthfully the heartbeats aren't really used anywhere yet, this was a thing we shipped as a proof of concept for a hackathon

> Should diagnostics.close_file() come after sys_diag_done_event.join()... there seems to be a possible race condition here.

Yeah good call. Right now it does this and we get a warning message sometimes saying we can't write to the Diagnostics file when we do our final cache flush to it. That could potentially be lost data, so we should close the file after we shut down the thread 👍

ametzger (approved):

jf_agent/main.py
herbal-jelly: Sounds good to me.

PR #344, opened 2024-07-11
[OJ-36936] Remove Agent Rollback Logic

jruel4:
Remove agent rollback logic - unstable.


ametzger (approved):
is it worthwhile to keep the timeout thing here? I'd almost be inclined to remove that as well, as it feels more like a foot-gun than a real utility IMO

PR #345, opened 2024-07-17
OJ-36861-setup-jf-ingest-for-github

gavinpitt-jf:
### Description

Set up JF Ingest Github GQL for all companies that are already running the original Agent version of the GQL ingest. I tested this with orthogonal-networks in an "end to end" test; I ran an Agent run for our git instance, imported it to `stg-looking-glass` via `git_import`, and then ran the Agent again. I verified that the first Agent run and submission created a few PRs and commits and branches, and then verified that the second run didn't create anything new. 

This PR contains initial groundwork that will be helpful as we roll out ADO in the Agent (via `jf_ingest`)

### DIFF
--- a/jf_agent/config_file_reader.py
+++ b/jf_agent/config_file_reader.py
@@ -11,7 +11,13 @@ import yaml
 from jf_agent import JELLYFISH_API_BASE, VALID_RUN_MODES
 from jf_agent.exception import BadConfigException
 from jf_ingest import logging_helper
-from jf_ingest.config import IngestionConfig, IngestionType, IssueMetadata, JiraDownloadConfig
+from jf_ingest.config import (
+    IngestionConfig,
+    IngestionType,
+    IssueMetadata,
+    JiraDownloadConfig,
+    GitConfig as JFIngestGitConfig,
+)
 
 from jf_agent.util import get_company_info
 
@@ -308,7 +314,9 @@ def _get_git_config_from_yaml(yaml_config) -> List[GitConfig]:
 git_providers = ['bitbucket_server', 'bitbucket_cloud', 'github', 'gitlab']
 
 
-def get_ingest_config(config: ValidatedConfig, creds, endpoint_jira_info: dict) -> IngestionConfig:
+def get_ingest_config(
+    config: ValidatedConfig, creds, endpoint_jira_info: dict, endpoint_git_instances_info: dict
+) -> IngestionConfig:
     """
     Handles converting our agent config to the jf_ingest IngestionConfig
     shared dataclass.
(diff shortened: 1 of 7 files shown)


jf_agent/config_file_reader.py
```diff
@@ -308,12 +315,50 @@ def _get_git_config_from_yaml(yaml_config) -> List[GitConfig]:
...
+                token=git_creds['github_token'],
+                base_url=config.git_url,
+                verify=not skip_ssl_verification,
+            )
+        if config.git_provider == GL_PROVIDER:
+            return None
+
+    except Exception as e:
+        logging_helper.log_standard_error(
+            logging.ERROR, msg_args=[config.git_provider, e], error_code=2101, exc_info=True,
```
andra-jellyfish, line 342: What does this error code map to?

jf_agent/config_file_reader.py
```diff
@@ -380,6 +425,79 @@ def get_ingest_config(config: ValidatedConfig, creds, endpoint_jira_info: dict)
...
+        pull_from = _make_datetimes_timezone_aware(endpoint_git_instance_info['pull_from'])
+        git_configs.append(
+            JFIngestGitConfig(
+                company_slug=company_slug,
+                instance_slug=instance_slug,
+                instance_file_key=endpoint_git_instance_info['key'],
+                git_provider=agent_git_config.git_provider,
+                git_auth_config=jf_ingest_git_auth_config,
+                url=agent_git_config.git_url,
+                jf_options=jf_options,
```
andra-jellyfish, line 486: is this a typo?

jf_agent/config_file_reader.py
gavinpitt-jf: This maps to "Failed to Connect", the lookup table is in the `logging_helper.py` [file in JF Ingest](https://github.com/Jellyfish-AI/jf_ingest/blob/491889ffeddc1b01aa95930bd6863e9951af64ae/jf_ingest/logging_helper.py#L101).

These Error codes were a part of JF Agent a LONG time ago, and as far as I know they aren't used meaningfully anywhere. Infact I think they do more harm than good by obscuring the code for no benefit. They're around as a legacy thing and should probably get revamped or removed

jf_agent/config_file_reader.py
```diff
@@ -380,6 +425,79 @@ def get_ingest_config(config: ValidatedConfig, creds, endpoint_jira_info: dict)
...
+        agent_git_config: GitConfig = agent_git_config
+
+        if is_mult_git_mode:
+            instance_slug = agent_git_config.git_instance_slug
+            endpoint_git_instance_info = endpoint_git_instances_info.get(instance_slug)
+            instance_creds = creds.git_instance_to_creds.get(instance_slug)
+        else:
+            # If there's only one git config set, then this is "single git" mode.
+            # The instance slug is likely not to be provided
+            endpoint_git_instance_info = list(endpoint_git_instances_info.values())[0]
```
jruel4, line 440: Q: Is this a result of having a "legacy" mode of sending things (non-GQL/ADO), as well as the new code? Are there any failure modes we're concerned about here?

jf_agent/git/utils.py
```diff
@@ -4,6 +4,17 @@
...
+'''
+
+    Constants
+
+'''
+BBS_PROVIDER = 'bitbucket_server'
+BBC_PROVIDER = 'bitbucket_cloud'
+GH_PROVIDER = 'github'
+GL_PROVIDER = 'gitlab'
+PROVIDERS = [GL_PROVIDER, GH_PROVIDER, BBS_PROVIDER, BBC_PROVIDER]
```
jruel4, line 16: Nit: Probably better do a tuple here to keep immutable unless we change this.

jf_agent/config_file_reader.py
jruel4: Or, now re-reading, is this not a requirement if the custom only provides a single git instance in their `config.yaml`?

jruel4 (approved):
LGTM pending a couple of questions

jf_agent/config_file_reader.py
gavinpitt-jf: This logic is used in a few places around the Agent: https://github.com/Jellyfish-AI/jf_agent/blob/08cfab5d49cd7f69d743519184aa9c31a8553b8f/jf_agent/main.py#L723

PR #346, opened 2024-07-25
Structlog dependency

lheureuxe13:
## Description

Adds a dependency for the `structlog` package to pdm. Also updates the lock strategy used by pdm.


acicchetti-jf (approved):
seems right to me!

gavinpitt-jf (approved):
If it seems right to Anthony it seems right to me 🔥

PR #353, opened 2024-07-30
OJ-37114 - Configure structlog

lheureuxe13:
## Description

Configures structlog for use in this repository. In the current configuration, structlog will handle log message rendering and formatting before sending a finalized log message to the standard `logging` library. The next step of utilizing structlog will be replacing all instances of the standard library logger with the new structlog logger + slightly modifying the `CustomQueueHandler` based on the new logging formatting.

Example log:

```{"event": "Structlog and standard logging configured", "agent_run_uuid": "8c2c7a73-3f9f-488e-9f1c-50e274fc8ce2", "jf_meta": {"commit": "bac1462befc405d8379160790f4085ca1732c6fe", "timestamp": "20240729T144304Z"}, "upload_time": "20240730_163839", "run_mode": "download_only", "company_slug": "orthogonal-networks", "logger": "__main__", "level": "info", "timestamp": "2024-07-30T16:38:39.569866Z", "lineno": 167, "filename": "main.py", "func_name": "main"}```

### DIFF
--- a/jf_agent/agent_logging.py
+++ b/jf_agent/agent_logging.py
@@ -2,12 +2,14 @@ from datetime import datetime, timedelta
 import logging
 from logging import LogRecord
 import os
+import structlog
 import sys
 import json
+import uuid
 from logging.handlers import QueueHandler, QueueListener
 from queue import Queue
 from dataclasses import dataclass
-from typing import Any, List, Union
+from typing import Any, List, Optional, Union
 from http.client import HTTPConnection, HTTPSConnection
 from urllib.parse import urlparse
 
@@ -29,6 +31,7 @@ e.g., function names, iteration counts
 '''
 
 LOG_FILE_NAME = 'jf_agent.log'
+LOG_FORMAT = logging.Formatter(fmt='%(message)s')
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: jf_agent/agent_logging.py [hunk 1]
```diff
@@ -217,6 +219,50 @@ def filter(self, record: LogRecord) -> bool:
...
+            structlog.processors.JSONRenderer(),
+        ],
+        wrapper_class=structlog.stdlib.BoundLogger,
+        logger_factory=structlog.stdlib.LoggerFactory(),
+        cache_logger_on_first_use=True,
+    )
+
+
+def bind_context_vars(
+    run_mode: str, company_slug: Optional[str], upload_time: str,
```
lheureuxe13, line 254: Type hinting `company_slug` as Optional since the value passed to that param comes from `.get()`

jruel4 (approved):
Question - what does the logging output look like that's printed out in console? We'll want to have structured logging going to AWS/S3, but human-readable log formatting printing out to the console.

jruel4 (changes requested):
Ahh didn't mean to approve, sorry

lheureuxe13:
@jruel4 The console output will look like this:

Screenshot 2024-07-30 at 3 29 17 PM

Would you prefer the console output not to include context vars? I should be able to update the config so that they're only excluded for the console

jf_agent/agent_logging.py
```diff
@@ -217,6 +219,53 @@ def filter(self, record: LogRecord) -> bool:
...
+            ),
+            structlog.dev.ConsoleRenderer(),
+        ],
+        wrapper_class=structlog.stdlib.BoundLogger,
+        logger_factory=structlog.stdlib.LoggerFactory(),
+        cache_logger_on_first_use=True,
+    )
+
+
+def bind_context_vars(
```
jruel4, line 254: Nit: This is somewhat confusingly similar to the structlog `bind_contextvars`, but with different functionality. Might change to something like `bind_default_agent_context` or similar

lheureuxe13: Good point, I can update the function name

jf_agent/agent_logging.py
```diff
@@ -217,6 +219,53 @@ def filter(self, record: LogRecord) -> bool:
...
+    company_slug: Optional[str],
+    upload_time: str,
+) -> None:
+    structlog.contextvars.clear_contextvars()
+    structlog.contextvars.bind_contextvars(
+        run_mode=run_mode,
+        company_slug=company_slug,
+        upload_time=upload_time,
+        agent_run_uuid=str(uuid.uuid4()),
+        jf_meta={'commit': os.getenv("SHA"), 'timestamp': os.getenv("BUILDTIME")},
```
jruel4, line 265: Nit: `timestamp` isn't readily apparent as being the build time here; consider renaming to `commit_build_time` or something similar

jf_agent/main.py
```diff
@@ -47,6 +47,9 @@
 from jf_ingest.jf_jira import load_and_push_jira_to_s3
 from jf_ingest.config import IngestionType, IngestionConfig
 
+# Configure structlog before any loggers are created
+agent_logging.configure_structlog()
```
jruel4, line 51: Q: Do we want structlogging from any non-`main.py` entrypoints? If so, consider initializing in the agent logging file.

This might be the right way to do things, just food for thought here.

jruel4:
> @jruel4 The console output will look like this:
> Screenshot 2024-07-30 at 3 29 17 PM
> 
> Would you prefer the console output not to include context vars? I should be able to update the config so that they're only excluded for the console

We do have some customers that parse through the logfile if something is awry. I think having the full context vars dumping into the console would make things hard to read. I'd say formatting the console output to be human-readable here would be preferable.

jf_agent/agent_logging.py
```diff
@@ -189,7 +199,7 @@ def handle(self, record: Union[logging.LogRecord, int]) -> None:
 
 @dataclass
 class AgentLoggingConfig:
-    level: str
+    level: int
     datefmt: str
     handlers: List[str]
     listener: QueueListener
```
lheureuxe13, line 205: Updating this since logging level is a int rather than a string. Also, we might want to reconsider if this dataclass is actually necessary

jf_agent/agent_logging.py
```diff
@@ -273,16 +397,21 @@ def print_to_log(*args):
     )
 
     logging.basicConfig(
-        level=config.level, datefmt=config.datefmt, handlers=config.handlers,
+        level=config.level,
+        datefmt=config.datefmt,
+        handlers=[logfile_handler, console_log_handler, log_queue_handler],
```
lheureuxe13, line 402: Changed this to the list of handlers, there was an issue parsing the handlers from the dataclass attribute

jf_agent/agent_logging.py
```diff
@@ -163,17 +191,28 @@ def handle(self, record: Union[logging.LogRecord, int]) -> None:
 
         if record != self._sentinel:
             msg = _standardize_log_msg(self.format(record))
-            self.messages_to_send.append(
-                {
-                    'message': msg,
-                    'timestamp': int(
+
+            # This logic is required to handle the different way that the timestamp attribute
```
jruel4, line 195: Q: Is there anything we could switch on here to determine which of these paths to take?

jf_agent/agent_logging.py
```diff
@@ -217,38 +256,146 @@ def filter(self, record: LogRecord) -> bool:
...
     log_queue_handler = CustomQueueHandler(log_queue, webhook_base, api_token)
-    log_queue_handler.setFormatter(file_and_queue_formatter)
+    log_queue_handler.setFormatter(json_formatter)
     log_queue_handler.setLevel(logging.DEBUG)
     queue_listener = CustomQueueListener(log_queue, log_queue_handler, respect_handler_level=True)
     queue_listener.start()
 
     # Silence the urllib3 logger to only emit WARNING level logs,
     # because the debug level logs are super noisy
+    logging.getLogger("urllib3").setLevel(logging.WARNING)
```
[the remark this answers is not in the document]
jruel4: @gavinpitt-jf do we want this? I remember something with Kirk where maybe we wanted to keep these logs?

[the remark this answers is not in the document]
gavinpitt-jf: I'd vote to change the comment and leave this as `DEBUG`

jf_agent/agent_logging.py
```diff
@@ -163,17 +191,28 @@ def handle(self, record: Union[logging.LogRecord, int]) -> None:
...
-                }
+                    )
+                    * 1000
+                )
+            except AttributeError:
+                created_time = datetime.fromtimestamp(record.created)
+                asctime = created_time.strftime('%Y-%m-%d %H:%M:%S,%f')[:-3]
+
+            timestamp = int(
+                datetime.strptime(asctime + '000', '%Y-%m-%d %H:%M:%S,%f').timestamp() * 1000
```
gavinpitt-jf, line 211: Ditto to what Jack said above, I'm also a little confused why were converting this to a string and then from a string back to a datetime?

I think you could go off of purely the `record.created` time, right? It comes back in already a timestamp formate (although we may still have to multiply it by 1000, I'm not sure about that)

https://docs.python.org/3/library/logging.html

jf_agent/agent_logging.py
```diff
@@ -1,16 +1,19 @@
...
-from queue import Queue
+import uuid
 from dataclasses import dataclass
-from typing import Any, List, Union
+from datetime import datetime, timedelta
 from http.client import HTTPConnection, HTTPSConnection
+from logging import LogRecord
+from logging.handlers import QueueHandler, QueueListener
+from queue import Queue
+from typing import Any, List, Optional, Union
```
lheureuxe13, line 12: Import rearrange due to isort

jf_agent/agent_logging.py
lheureuxe13: Simplified this to always use `.created`

jf_agent/agent_logging.py
lheureuxe13: Agreed, using `.created` is much simpler, I've updated the code

jf_agent/agent_logging.py
```diff
@@ -217,38 +246,157 @@ def filter(self, record: LogRecord) -> bool:
...
     outdir: str, webhook_base: str, api_token: str, debug_requests=False
 ) -> AgentLoggingConfig:
+    # Remove default handlers that are added when logging is used before configuring
+    logging.getLogger().handlers.clear()
+
+    # Configure structlog before the standard logging module
+    configure_structlog()
+
+    # Set an envvar so tdqm works with structlog
+    os.environ["TDQM_POSITION"] = "-1"
```
jruel4, line 376: Nice!! cc @gavinpitt-jf might be useful elsewhere? https://tqdm.github.io/releases/#v4661-2023-08-10

jruel4 (approved):

PR #382, opened 2024-12-30
[OJ-40908] Only send diagnostic data if JF webhooks endpoint is reachable

lheureuxe13:
Only try to send agent diagnostic data to the JF webhooks endpoint if the earlier test to send logs to the same domain was successful


gavinpitt-jf (approved):

andra-jellyfish (approved):
lgtm

PR #383, opened 2024-12-30
[OJ-41016] JQL filter

lheureuxe13:
Add JQL filter to config object passed to `jf_ingest` and bump `jf_ingest` version

### DIFF
--- a/jf_agent/config_file_reader.py
+++ b/jf_agent/config_file_reader.py
@@ -454,6 +454,7 @@ def get_ingest_config(
             only_issues=False,
             recursively_download_parents=config.jira_recursively_download_parents,
             jellyfish_issue_ids_for_redownload=jf_issue_ids_for_redownload,
+            issue_jql_filter=config.jira_issue_jql,
             #
             # worklogs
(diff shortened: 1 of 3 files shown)


jf_agent/config_file_reader.py
```diff
@@ -454,6 +454,7 @@ def get_ingest_config(
             only_issues=False,
             recursively_download_parents=config.jira_recursively_download_parents,
             jellyfish_issue_ids_for_redownload=jf_issue_ids_for_redownload,
+            issue_jql_filter=config.jira_issue_jql,
```
gavinpitt-jf, line 457: What happens if this is not provided by the config.yml? Is this value null, or `''`? I think we would want it to be `None` if it's not provided, right? that's how it used to work (the default value if not provided is `None` isn't it?)

gavinpitt-jf (approved):

jf_agent/config_file_reader.py
lheureuxe13: That's a good point, I've set the default to `None` rather than `''` when the values are read in from the config file

PR #388, opened 2025-01-09
OJ-41290 Update GitLab list pulls to use iterator keyword

ashleyredzko:
This updates our `.list()` calls that were previously passing `as_list=False` to get a generator back to instead pass `iterator=True`. This was a breaking change introduced with newer versions of the GitLab dependency used to call the API.


yashayang2024 (approved):
Looks good to me. Thanks for the hotfix!

PR #394, opened 2025-03-25
OJ-42967 Support JF Ingest Gitlab adapter

ashleyredzko:
Adds support for using the JF Ingest adapter for GitLab. Includes an ingest version bump.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -25,7 +25,7 @@ jobs:
         python-version: "3.10"
     - name: Install dependencies
       run: |
-        pip install pdm && pdm install --dev
+        sudo apt-get install libkrb5-dev && pip install pdm && pdm install --dev
     - name: Tests (pytest)
       run: |
         pdm run python -m pytest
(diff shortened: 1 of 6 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         python-version: "3.10"
     - name: Install dependencies
       run: |
-        pip install pdm && pdm install --dev
+        sudo apt-get install libkrb5-dev && pip install pdm && pdm install --dev
```
jruel4, line 28: Q: What's the reason for this?

ashleyredzko: Because of the updated required packages for the ingest bump, this action was running into issues (Failed run here). This fix was taken after stumbling across [this issue](https://github.com/jupyter-incubator/sparkmagic/issues/762) and [this stack overflow answer](https://stackoverflow.com/a/50410443/4978266).

I'm not sure if this needs to be included in the dockerfile or not, but this fixed the action.

ashleyredzko: Was able to run `docker build .` locally and the image successfully built, so this is a workaround for the action, which has issues with symlinking krb5 for gssapi.

jruel4 (approved):

PR #396, opened 2025-03-31
OJ-43768 Handle non-str values for included and excluded repos

ashleyredzko:
When we're generating the git configurations, we take in whatever values are present in the yaml file. However, we do some manipulation of the repository names in `jf_ingest` and they need to be strings for that. GitLab's repository names are ids, so they are ingested as integers. Since we expect this everywhere, cast each value to str when we're building the config object instance.


jruel4 (approved):

gavinpitt-jf (changes requested):
I think this will break for non-JF Ingest customers. This is changing the type at the config file level reader, but we likely only want to change the type right before we pass the values to JF Ingest!

ashleyredzko:
Moved this logic to when we're generating the `JFIngestGitConfig` per @gavinpitt-jf 's comment

gavinpitt-jf (approved):

PR #403, opened 2025-04-24
OJ-44311 Pass ssl verification to Gitlab library client

ashleyredzko:
The Gitlab client provided by the library that we use to access the api does not respect the session verification setting and has it's own value it looks for when attempting api calls.

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "dev"]
 strategy = ["inherit_metadata"]
 lock_version = "4.5.0"
-content_hash = "sha256:9149cdf4c5adb1e51650749e37c5938bcc82b766ff06771ba2c41da8de43a829"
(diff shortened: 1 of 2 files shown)


gavinpitt-jf (approved):
This looks good to me, but do we have to change any default values here like I mentioned we might need to for direct connect? Is the `verify` arg passed in as part of the config?

ashleyredzko:
@gavinpitt-jf Agent pulls this via `no_verify_ssl`, and defaults it to `False`, which would give us a `verify` of `True`. Should be good to avoid additional changes.
https://github.com/Jellyfish-AI/jf_agent/blob/63462aff078c768a594c3ac01d25dc7c2f18c530/jf_agent/config_file_reader.py#L141
https://github.com/Jellyfish-AI/jf_agent/blob/63462aff078c768a594c3ac01d25dc7c2f18c530/jf_agent/config_file_reader.py#L348

PR #450, opened 2026-04-29
OJ-54337 Use repo slug for bitbucket URLs

gunthercox:
**Relevant API Documentation:**

* https://developer.atlassian.com/cloud/bitbucket/rest/api-group-repositories/#api-group-repositories
* https://developer.atlassian.com/cloud/bitbucket/proxy-object-hydration/
* https://developer.atlassian.com/cloud/bitbucket/bitbucket-api-teams-deprecation/

### DIFF
--- a/jf_agent/git/bitbucket_cloud_adapter.py
+++ b/jf_agent/git/bitbucket_cloud_adapter.py
@@ -385,12 +385,12 @@ def _standardize_user(api_user):
 def _standardize_pr(
     client, repo, api_pr, strip_text_content: bool, redact_names_and_urls: bool,
 ):
-    slug = _repo_slug(repo)
+    repo_slug = _repo_slug(repo)
 
     # Process the PR's diff to get additions, deletions, changed_files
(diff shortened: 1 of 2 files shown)


ashleyredzko (approved):

jf_agent/git/bitbucket_cloud_adapter.py
ashleyredzko, line 167: Does this also need the slug?

jf_agent/git/bitbucket_cloud_adapter.py
```diff
@@ -210,7 +207,7 @@ def get_pull_requests(
                         server_git_instance_info, repo.project.login, repo.id, 'prs'
```
ashleyredzko, line 207: Checking here, should this use the slug as well?

gunthercox: I'll double check, but I believe this and the other line can both be kept as UUIDs. The goal is to keep the UUID usage internally, but ensure API requests are all made using the slug.

PR #452, opened 2026-05-07
OJ-54189 Enable using group names to pull Jira users

ashleyredzko:
This allows supplying a set of group names to use when we pull users. If not provided (or is empty), we fall back to `user/search`.


lheureuxe13 (approved):
LGTM (once jf_ingest has been updated)

gavinpitt-jf (approved):

PR #461, opened 2026-06-03
OJ-55279 Track commits that have occured more than once across branches

gunthercox:
The Bitbucket Server adapter yields duplicate commits when multiple branches are configured because the `/commits` endpoint returns all commits reachable from the branch tip (i.e. the commit itself plus every ancestor back through the parent chain), not just commits unique to that branch.

This can be fixed by tracking seen commit hashes per-repo and skipping duplicates. Since the default branch is always processed first, its commits are attributed there, and feature branches only yield their unique diverged commits. This is safe because commit-to-PR association is handled separately via PR metadata, not branch attribution.

### DIFF
--- a/jf_agent/git/bitbucket_server.py
+++ b/jf_agent/git/bitbucket_server.py
@@ -312,15 +312,15 @@ def get_commits_for_included_branches(
                                 tqdm.write(
                                     f"[{datetime.now().isoformat()}] Getting {commit['id']} ({repo['name']})"
                                 )
+                            seen_hashes.add(commit['id'])
                             standardized_commit = _standardize_commit(
                                 commit, repo, branch, strip_text_content, redact_names_and_urls
                             )
                             # commits are ordered newest to oldest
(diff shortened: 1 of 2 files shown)


lheureuxe13 (approved):

jf_agent/git/bitbucket_server.py
```diff
@@ -305,15 +306,18 @@ def get_commits_for_included_branches(
                         start=1,
                     ):
                         with logging_helper.log_loop_iters('branch commit inside repo', j, 100):
+                            if commit['id'] in seen_hashes:
+                                continue
```
ashleyredzko, line 310: Nit: If verbose is enabled, it might be nice to log that we're skipping this hash because it's already been seen. Probably not really needed though.

ashleyredzko (approved):

gunthercox:
Closing this off in preference to an upstream deduplication solution.

PR #467, opened 2026-08-14
CORE-221: Handle ChunkedEncodingError in PR activities and commits fetching

gavinpitt-jf:
> [!NOTE]
> This PR was agent-generated. Human testing and review required.
>
> — SHIPMATE

### Description

This PR adds retry logic for `ChunkedEncodingError` on the `api_pr.activities()` and `api_pr.commits()` calls in the Bitbucket Server adapter. Previously, a broken connection mid-chunked-transfer on these calls would propagate up and abort the entire git download (7+ hours of work for large customers like Keysight). Now the agent retries up to 3 times with exponential backoff before gracefully skipping the affected PR's data.

The `diff()` call already caught `ChunkedEncodingError` (skipping without retry), but activities and commits had no handling at all — a single transient network failure would kill the entire run at error code 3061.

### Testing

- [x] Agent validation: existing `test_bitbucket_server.py` tests pass (8/8)
- [ ] Manual testing needed:
  - [ ] Verify agent successfully retries and recovers from a transient connection drop during PR activities/commits pagination

### Rollback

- [x] PR revert (safe — no migrations or data changes)

### Notes

- Root cause was observed on the `keysight` customer agent on 2026-08-13: the agent hung for ~10 minutes on a chunked activities response for `pma-utils` PR #15 before the connection broke, killing the entire 7-hour run.
- Retry uses exponential backoff (1s, 2s, 4s) — light enough to not delay the overall run meaningfully, but enough to ride out transient network hiccups.
- After 3 failed attempts, behavior degrades gracefully: logs a warning and continues with empty activities/commits for that PR (same pattern as the existing `diff()` handler).

### DIFF
--- a/jf_agent/git/bitbucket_server.py
+++ b/jf_agent/git/bitbucket_server.py
@@ -1,4 +1,5 @@
 import logging
+import time
 from datetime import datetime
 
 import pytz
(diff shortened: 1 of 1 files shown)


gavinpitt-jf:
bugbot run

ashleyredzko (approved):

PR #470, opened 2026-09-03
CORE-336: Honor server-delivered git branch config

jsemeraro:
### Description

Read git branch selection from the `get_pull_state` payload in `get_ingest_config`, so branch config set on the server reaches jf_ingest.

`repo_id_to_pull_all_commits_and_branches` controls whether jf_ingest lists a repo's branches at all. Without it, `included_branches_by_repo` has nothing to narrow, since it is applied to the branches discovery already returned. The agent config file has no equivalent option, so previously `include_branches` in `config.yml` had no effect.

Server values win per repo. Repos absent from the server payload keep their local config. Each field is read only when its key is present, so an older server payload leaves current behavior unchanged.

Follows the pattern used for `backpopulation_window_days`.

Requires the paired server change.

### Testing Strategy
- [x] Local Validation

`tests/test_config_file.py` — 12 passed. Verified the three kwargs exist on `GitConfig` in the pinned `jf-ingest==0.0.308`.

### Rollback Strategy
- [x] PR Revert

### Concerns / Assumptions / Tradeoffs

This agent release is required for the server change to take effect. An older agent ignores the new payload keys silently.

Instances that already have all-branches enabled on the server will begin pulling every branch once this ships. Review those before release.

### DIFF
--- a/jf_agent/config_file_reader.py
+++ b/jf_agent/config_file_reader.py
@@ -571,6 +571,26 @@ def get_ingest_config(
                 'backpopulation_window_days'
             ]
 
+        # Branch selection configured in Jellyfish wins per repo; repos the server says
+        # nothing about keep whatever the local config file asked for. Without this, a
+        # branch configured in Jellyfish silently has no effect on an agent pull, so
+        # commits made directly to a non-default branch are never collected.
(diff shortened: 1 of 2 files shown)


jsemeraro:
bugbot run

ashleyredzko (approved):

